PR #46, opened 2015-01-23
Remove unsafe

anlhord:
Please review the remove unsafe change.

This change removes uses of unsafe for deploying an aerospike client to the google app engine.


khaf:
I will have a look and merge latest Monday.

PR #48, opened 2015-03-01
Fix: newBatchCommandGet() ignoring binNames restrictions

dademurphy:
At present, BatchGet(policy BasePolicy, keys *[]Key, bins ...string) ([]Record, error) altogether ignores the |bins| argument. Looks like it's omitted in newBatchCommandGet(), and with this CL I'm seeing restrictions being correctly applied.


khaf:
Hi,

Thank you for your pull request, and sorry for the late response. I'll have a look, add tests, and include your PR in the next release.

Cheers,
Khosrow

PR #50, opened 2015-03-05
Fix error returning on all commands

faizanchango:


command.go
```diff
@@ -117,8 +117,8 @@ func (cmd *baseCommand) setWrite(policy *WritePolicy, operation OperationType, k
 		cmd.writeFieldValue(key.userKey, KEY)
 	}
 
-	for i := range bins {
-		if err := cmd.writeOperationForBin(bins[i], operation); err != nil {
+	for _, bin := range bins {
```
khaf: We avoid this to avoid memory copies.

command.go
```diff
@@ -128,11 +128,11 @@ func (cmd *baseCommand) setWrite(policy *WritePolicy, operation OperationType, k
...
 
 // Writes the command for delete operations
-func (cmd *baseCommand) setDelete(policy *WritePolicy, key *Key) error {
+func (cmd *baseCommand) setDelete(policy *WritePolicy, key *Key) (err error) {
 	cmd.begin()
 	fieldCount := cmd.estimateKeySize(key)
-	if err := cmd.sizeBuffer(); err != nil {
-		return nil
+	if err = cmd.sizeBuffer(); err != nil {
+		return
```
khaf: Explicit returns are preferable.

read_command.go
```diff
@@ -88,10 +88,6 @@ func (cmd *readCommand) parseResult(ifc command, conn *Connection) error {
 	}
 
 	if resultCode != 0 {
-		if resultCode == KEY_NOT_FOUND_ERROR {
```
khaf: This a matter of policy in all clients. WE had to do the same in Go client.

faizanchango: Is there another way we can we figure out a key not found error if we return nil here?

khaf: When the `len(record.Bins) == 0`. Aerospike does not keep empty records.

GeertJohan:
I think this PR can be closed. It's really not preferable to use named return values and naked returns in functions/methods larger than just a few lines.

PR #59, opened 2015-05-10
Fixed error handling on authentication failure

imkira:
Even on authentication failure, no errors are returned due to improper error handling.
This PR is an attempt to fix it.
The same problem may also occur when calling SetTimeout, but since it is not currently possible to test that and the fix is similar, I am including in the same PR.

### DIFF
--- a/node.go
+++ b/node.go
@@ -220,14 +220,14 @@ L:
 		}
 
 		// need to authenticate
-		if conn.Authenticate(nd.cluster.user, nd.cluster.password); err != nil {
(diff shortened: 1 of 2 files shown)


khaf:
Found another one in `NodeValidator` and fixed it.

imkira:
The merge kinda went wrong it seems:

https://github.com/aerospike/aerospike-client-go/blob/master/node.go#L235-L246

khaf:
ARGH... and more than just the error.

khaf:
Fixed.

PR #61, opened 2015-05-17
Fixed invalid connection handling on node connections

rndive:
I realized that Node connectionCount is not decrement when connections don't return through PutConnection.

There is a possibility to block new connection if LimitConnectionsToQueueSize is true and connections have some problems.

This PR is an attempt to fix it.

I'm not sure whether the connection pool handling on Node.Refresh is correct or not...


khaf:
Thanks for the PR. Will merge on next release.

PR #69, opened 2015-06-22
Fix a bunch of golint notices.

GeertJohan:
*These changes do not modify any behavior.*

I noticed that golint was giving a lot of warnings/notices, so I took a few seconds to get rid of some of them. Most changes are correcting comment formats and removing of unrequired else blocks.

I'm ignoring some notices (`grep -v`). The first (dot import) is a lot of work to get rid of, and not that important to the user. The second (caps) and third (Id>ID) would require breaking changes.
These golint notices are now left:
```
$ golint | grep -v "should not use dot imports" | grep -v "don't use ALL_CAPS in Go names" | grep -v "Id should be"
admin_command.go:57:6: exported type AdminCommand should have comment or be unexported
marshal.go:157:6: exported type SyncMap should have comment or be unexported
recordset.go:23:6: exported type Result should have comment or be unexported
role.go:18:6: exported type Role should have comment or be unexported
value.go:61:6: type name will be used as aerospike.AerospikeBlob by other packages, and that stutters; consider calling this Blob
```

### DIFF
--- a/operation.go
+++ b/operation.go
@@ -17,9 +17,12 @@ package aerospike
 // OperationType determines operation type
 type OperationType byte
 
+// Valid OperationType values that can be used to create custom Operations.
+// The names are self-explanatory.
 const (
 	READ OperationType = 1
 	// READ_HEADER OperationType = 1
+
 	WRITE   OperationType = 2
 	ADD     OperationType = 5
 	APPEND  OperationType = 9
--- a/admin_policy.go
+++ b/admin_policy.go
@@ -16,7 +16,7 @@ package aerospike
 
 import "time"
 
-// Policy attributes used for user administration commands.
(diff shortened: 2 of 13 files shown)


operation.go
```diff
@@ -17,9 +17,12 @@ package aerospike
 // OperationType determines operation type
 type OperationType byte
 
+// Valid OperationType values that can be used to create custom Operations.
+// The names are self-explanatory.
 const (
 	READ OperationType = 1
 	// READ_HEADER OperationType = 1
+
 	WRITE   OperationType = 2
```
GeertJohan: This newline here might look strange, but without it the `// READ_HEADER …(etc)` comment becomes documentation for `WRITE`.

PR #71, opened 2015-07-02
Fix *Connection.Read() total bytes count on error

GeertJohan:
I noticed that the total read bytes count returned `*Connection.Read()` could be off from the actual number of read bytes when an error occurred. In some cases a net.Conn can read bytes (n>0) while also returning with an error (for example: EOF).


khaf:
Good catch, thanks.

GeertJohan:
@khaf can you merge this (and other open PR's) into master?

khaf:
@GeertJohan I'll definitely do it tomorrow. I'll need to read some other cases in the code to make sure everything is working.

PR #76, opened 2015-08-08
Fixed typo.

charl:
Fixed typo.


GeertJohan:
I think this can be merged, right?

khaf:
Should have been merged. There's a glitch somewhere that doesn't work as expected.

PR #81, opened 2015-09-11
Use IP map when adding friends, similar to the java library

chrisguiney:
more or less taken from the java library.   Please feel free to let me know if there's anything else that would be required for this.


khaf:
Thanks for the PR. Merged if the latest release.

PR #83, opened 2015-09-22
Fix PackLong, fix #78

### LINKED ISSUE #78: Either packer.PackLong or packer.PackULong is broken
[`PackLong`](https://github.com/aerospike/aerospike-client-go/blob/652f5ca4caecf34d29fd78eacbf879f2cefa2568/packer.go#L343) Looks like:
```go
func (pckr *packer) PackLong(valType int, val int64) {
	pckr.buffer.WriteByte(byte(valType))
	pos := pckr.grow(_b8)
	pckr.buffer.Write(Buffer.Int64ToBytes(val, pckr.buffer.Bytes(), pos))
}
```

While [`PackULong`](https://github.com/aerospike/aerospike-client-go/blob/652f5ca4caecf34d29fd78eacbf879f2cefa2568/packer.go#L349) looks like:
```go
func (pckr *packer) PackULong(val uint64) {
	pckr.buffer.WriteByte(byte(0xcf))
	pos := pckr.grow(_b8)
	Buffer.Int64ToBytes(int64(val), pckr.buffer.Bytes(), pos)
}
```
Is the `pckr.buffer.Write` in `PackLong` needed or is it also needed in `PackULong`?

ErikDubbelboer:


khaf:
Thanks for the PR. Merged if the latest release.

PR #105, opened 2015-12-08
Fix comments and cleanup Geo filter functions

GeertJohan:
Just some minor changes.

### DIFF
--- a/filter.go
+++ b/filter.go
@@ -15,7 +15,6 @@
 package aerospike
 
 import (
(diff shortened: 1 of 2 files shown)


khaf:
This breaks the GeoJson protocol. I'll take care of the comment fixes later, but the code was sound and as it should have been.

GeertJohan:
Apologies. I was under the impression that my changes were functionally the same, just a bit cleaner.

PR #124, opened 2016-03-28
vet and gofmt

alicebob:
Fix some `go vet` warnings (and run `gofmt` while we're at it).

### DIFF
--- a/client.go
+++ b/client.go
@@ -1021,7 +1021,6 @@ func (clnt *Client) QueryAggregate(policy *QueryPolicy, statement *Statement, pa
 		},
 			luaArgs...,
 		); err != nil {
-			panic(err)
 			recSet.Errors <- err
 			return
 		}
--- a/client_object_test.go
+++ b/client_object_test.go
@@ -535,7 +535,6 @@ var _ = Describe("Aerospike", func() {
 				}
 
 				for e := range rs.Errors {
-					panic(e.Error())
 					Expect(e).ToNot(HaveOccurred())
 				}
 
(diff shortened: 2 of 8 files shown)


client.go
```diff
@@ -1022,8 +1022,6 @@ func (clnt *Client) QueryAggregate(policy *QueryPolicy, statement *Statement, pa
 			luaArgs...,
 		); err != nil {
 			panic(err)
```
khaf: It is the panic which should be removed.

client_object_test.go
```diff
@@ -536,7 +536,6 @@ var _ = Describe("Aerospike", func() {
 
 				for e := range rs.Errors {
 					panic(e.Error())
```
khaf: Same here!

client_object_test.go
```diff
@@ -578,7 +577,6 @@ var _ = Describe("Aerospike", func() {
...
```
khaf: And here!

pkg/bcrypt/cipher.go
```diff
@@ -272,10 +272,9 @@ var bf_crypt_ciphertext = [6]uint{
 	0x64657253, 0x63727944, 0x6f756274,
 }
 
-
 type cipher struct {
-	P [18]uint
```
khaf: This package is external; I'd rather not modify it.

khaf:
Thanks for the PR.

Looks good. If you remove the panics and return the original code, I'll merge it in.

alicebob:
Applied your comments. It's a push -f to clean up the panic() removals.

khaf:
Merged. Thanks.

PR #126, opened 2016-04-02
fix locked goroutines

yiiton:


cluster.go
```diff
@@ -229,7 +229,7 @@ func (clstr *Cluster) tend() error {
 func (clstr *Cluster) waitTillStabilized() {
 	count := -1
 
-	doneCh := make(chan bool)
+	doneCh := make(chan bool, 1)
```
khaf: These two are by definition equivalent. What is the issue that you're trying to solve by this change?

erikdubbelboer: They are not the same https://play.golang.org/p/86islc9_lh runs fine while https://play.golang.org/p/feLEyinaIE results in a deadlock.

[here](https://github.com/aerospike/aerospike-client-go/blob/ad344291c1d9eadc47fc9f2b913449a4598a4bfe/cluster.go#L251) the gorountine tries to put `true` in. If this happens after the timeout below nobody will be reading from the channel and the gorountine will hang forever trying to put it in. Setting the buffer to 1 allows the goroutine to always put `true` in and return while not modifying the overall behavior.

khaf: uh oh, you're right. I stand corrected. Will merge in next release.

khaf:
New version released. Thanks.

PR #127, opened 2016-05-12
Run go fmt ./...

erikdubbelboer:
See https://github.com/jameskeane/bcrypt/pull/4


erikdubbelboer:
Those failures have nothing to do with this cosmetic change. Seems aerospike doesn't want to start on travis anymore.

khaf:
There's no problem with your change. I'll merge it in for next release. Thanks.

PR #128, opened 2016-05-13
recordset: try to make r.Close kill r.Results chan

nilium:
This attempts to fix a weakness in the current Recordset method set
(lack of mostly synchronous read) and a goroutine leak caused by
a forever-blocking send on the channel returned by Results(). The
blocking channel is only closed if and only if the entire record set is
read to completion despite it being close-able.

Currently, the results channel appears to be held open on sends because
Close() hits the cancellation channel but the Results() channel ignores
it, so it attempts to send on a channel that may have stopped being
received on and hangs.

This attempts to address that in two ways:

1. Add a synchronous Read method that does away with the channel,
   allowing consumers of the API to model concurrent access more
   without using the two to-be-unexported Records and Errors channels
   and without assuming that the one and only concurrent reader
   provided by the client library is adequate (since the current
   behavior of leaking goroutines after closing is bad).  This could
   obviously be renamed, but it seemed most appropriate to read
   a record from a record set.

2. Require all sends and receives from Records and Errors inside of the
   Results() reader goroutine to also receive on the cancelled channel,
   allowing the goroutine to bail out early if need be. Unlike Read(),
   this will not return an error (ErrRecordsetClosed) if the recordset
   is closed.

Close now returns an error if and only if it has been called more than
once. This is still safe to call more than once, but it provides an
indication that the record set received a double-close, which is useful
to know when debugging.

Previously, there were flush loops in the goroutine for handling what
I assume is an odd attempt to clean out a channel that will be GC'd
(along with its contents) after being closed in the Close() method.
This does not account for any other goroutines that may be holding the
waitgroup open far longer than needed. These flush loops have been
removed to not waste time on memory that will be scrapped.

### DIFF
--- a/recordset.go
+++ b/recordset.go
@@ -15,6 +15,7 @@
 package aerospike
 
 import (
+	"errors"
 	"fmt"
 	"reflect"
 	"sync"
@@ -102,6 +103,24 @@ func (rcs *Recordset) IsActive() bool {
 	return rcs.active.Get()
 }
 
+var ErrRecordsetClosed = errors.New("aerospike: Recordset closed")
+
+// Read reads the next record from the Recordset. If the Recordset has been
+// closed, it returns ErrRecordsetClosed.
+func (rcs *Recordset) Read() (record *Record, err error) {
+	var ok bool
+	select {
+	case <-rcs.cancelled:
+	case record, ok = <-rcs.Records:
+	case err, ok = <-rcs.Errors:
+	}
+
+	if !ok {
+		err = ErrRecordsetClosed
+	}
+	return record, err
+}
+
 // Results returns a new receive-only channel with the results of the Scan/Query.
 // This is a more idiomatic approach to the iterator pattern in getting the
 // results back from the recordset, and doesn't require the user to write the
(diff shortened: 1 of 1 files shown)


nilium:
Submitting this mostly so I can get another set of eyeballs on it, as we've noticed that partial reads of a Recordset result in a goroutine leak. This is caused, we think, by the Results method's goroutine reading a record and trying to write it back over its output channel, but since the reader has since moved on and closed the Recordset, it just hangs open.

recordset.go
```diff
@@ -102,6 +103,24 @@ func (rcs *Recordset) IsActive() bool {
...
+var ErrRecordsetClosed = errors.New("aerospike: Recordset closed")
+
+// Read reads the next record from the Recordset. If the Recordset has been
+// closed, it returns ErrRecordsetClosed.
+func (rcs *Recordset) Read() (record *Record, err error) {
+	var ok bool
+	select {
+	case <-rcs.cancelled:
+	case record, ok = <-rcs.Records:
+	case err, ok = <-rcs.Errors:
```
khaf: Thank you for your suggestions. You are correct that there is a potential goroutine leak in that method (it can be worse since the resources for the command won't be freed as well).

Interestingly, the problem is not as simple as it seems. For example, here you should not check if the error channel is closed, since if both chans are closed (happens in `Close` method), some records may be left in the `Records` chan. Please take a look at the following example to see how ```close(chan)``` works: https://play.golang.org/p/XRnLcmxeBL
There's also no need to directly read `cancelled` channel since it will automatically close the `Records` and `Errors` chans in time from the worker commands.

More troubling is that there's no way of automatically deallocating the resources in case the user forgets to do so, without consuming all records or manually calling the ```Close``` method. 

There's also the issue of the error type for the end of recordset which is not compatible with the generic ```AerospikeError```.

I see your point though, and will take care of the required changes for the next release, including introduction of the new ```Read()``` method.

ericdmann:
Thanks khaf -- do we have an expected date for the next release? With no way of limiting db side, I'm left with open channels or a memory problem. Seems I'm always forced to read in all results -- and still when calling Close, I'm seeing the open resources still laying around.

khaf:
Monday Latest for this one :)

Though you should always defer `rs.Close()` even with the new release.

Let me know in case there is something else I could somehow squeeze into the new release.

ericdmann:
Thanks awesome -- will update and let you know what I find.

recordset.go
nilium: It's fine if there are records left in either chan. After Close is called, we can assume the recordset will be discarded, the channels GC'd, and with it the records (or errors) in them. So, the channels will be closed, the recordset GC'd, and the world will move on just fine.

Checking for cancelled being closed is the easiest way to determine if the recordset is closed since it's a direct side-effect of Close being called, before any of the Records or Errors channels are closed. As such, it's also the easiest way to immediately bail out of reading off the records/errors channels.

The resources are automatically deallocated by calling Close and allowing the garbage collector to do its job. If the Aerospike client is retaining the same resources elsewhere beyond the recordset's lifetime, I would say that needs to be fixed: once the recordset is closed, that operation is done, the channels should be discarded, and the recordset is no longer valid for use.

khaf: You are forgetting that `Close()` is always called after the producer commands are done (hence the wait group). So if someone is consuming the recordset, and all the producers finish and `Close()` is called, some records may remain unused in the `Records` chan.

nilium: And that's fine. They called Close – they already don't care what they're discarding in your buffered channel. Why does it matter if we throw records away after we've closed the record set?

khaf: The recordset is closed automatically from the producer threads as well. We are not talking about cancelling, but the main use-case of consuming all records.

nilium: Me and Eric are talking about cancelling — i.e., closing the record set before it has been read to completion.

khaf: And I'm saying the same mechanism is being used for normal consumption, and your code to fix cancellation breaks that other use case :)

As I said, not as simple as it seems. I've already fixed it, and will release the fix today.

nilium: In that case, I think I'm just obviously misunderstanding you. I'll go ahead and close this pull request for the time being.

khaf: No problem. I've pulled this PR in the code, and changed it a bit. Thanks for your help and thoughts.

khaf:
New version released. Thanks for your help.

PR #132, opened 2016-06-09
Reverted KeepConnection to only discard connection errors

zachbadgett:
Not sure why this was changed however I would think you would want to only invalidate the connection on a connection error.

### DIFF
--- a/types/result_code.go
+++ b/types/result_code.go
@@ -244,11 +244,24 @@ func KeepConnection(err error) bool {
 	}
 
 	switch ae.resultCode {
-	case KEY_NOT_FOUND_ERROR:
-		return true
+	case 0, // Zero Value
+		QUERY_TERMINATED,
+		SCAN_TERMINATED,
+		INVALID_NODE_ERROR,
+		PARSE_ERROR,
+		SERIALIZE_ERROR,
+		SERVER_MEM_ERROR,
+		TIMEOUT,
+		SERVER_NOT_AVAILABLE,
+		SCAN_ABORT,
+		INDEX_OOM,
+		QUERY_ABORTED,
+		QUERY_TIMEOUT:
+
+		return false
 
 	default:
-		return false
+		return true
 	}
 }
 


khaf:
We had some Buffer Poisoning due to this, and I had to revert until I get a good handle on what caused those.

Many of the errors you have there in the list also leave the connection buffer intact, which will definitely poison the buffer.

zachbadgett:
Maybe I'm not understanding completely. Currently, on the master branch, the go client closes all connections if an error occurs except for KEY_NOT_FOUND_ERROR, wouldn't this cause the same issue? or did you mean the blacklist could be missing errors that cause the problem?

khaf:
Sorry for my late reply, the notification email vanished under a mountain of others :)

Some errors are returned from the server, and some others are client-side. Depending on the type of the error, we may invalidate the connection just in case the buffer has been poisoned. At the moment, we do this for most errors other than `KEY_NOT_FOUND_ERROR` which comes from the server and is fully parsed.

zachbadgett:
That makes more sense now. Would adding to the whitelist be acceptable then?

khaf:
Absolutely! I'm going to think about this for the next release, since adding to the white list will require some thorough testing.

absolute8511:
We saw the same problem recently which will cause unnecessary close of connection. Actually, I  think we can handle the error in the way of c client and Java client.

```
			switch (status) {
				case AEROSPIKE_ERR_TIMEOUT:
					as_node_close_connection(node, fd);
					if (release_node) {
						as_node_release(node);
					}
					return as_error_update(err, AEROSPIKE_ERR_TIMEOUT,
							"Timeout: timeout=%d iterations=%u failedNodes=%u failedConns=%u",
							timeout_ms, ++iterations, failed_nodes, failed_conns);
				
				// Close socket on errors that can leave unread data in socket.
				case AEROSPIKE_ERR_QUERY_ABORTED:
				case AEROSPIKE_ERR_SCAN_ABORTED:
				case AEROSPIKE_ERR_CLIENT_ABORT:
				case AEROSPIKE_ERR_CLIENT:
					as_node_close_connection(node, fd);
					if (release_node) {
						as_node_release(node);
					}
					err->code = status;
					return status;
				
				default:
					err->code = status;
					break;
			}
```

```
	public static boolean keepConnection(int resultCode) {
		// Keep connection on TIMEOUT because it can be server response which does not 
		// close socket.  Also, client timeout code path does not call this method. 
		switch (resultCode) {
		case 0: // Exception did not originate on server.
		case QUERY_TERMINATED:
		case SCAN_TERMINATED:
		case PARSE_ERROR:
		case SERIALIZE_ERROR:
		case SERVER_NOT_AVAILABLE:
		case SCAN_ABORT:
		case QUERY_ABORTED:
			return false;
			
		default:
			return true;					
		}
	}
```

khaf:
The latest release takes care of this. Please let me know in case you notice an issue.

zachbadgett:
Awesome!

PR #139, opened 2016-07-18
Don't use math.MaxInt64, it breaks 32-bit builds

sparrc:
before change:

$ GOOS=linux GOARCH=386 go build .
./value.go:103: constant 9223372036854775807 overflows uint

after change:

$ GOOS=linux GOARCH=386 go build .

### DIFF
--- a/value.go
+++ b/value.go
@@ -18,7 +18,6 @@ import (
 	"bytes"
 	"fmt"
 	"io"
-	"math"
 	"reflect"
 	"strconv"
 	"strings"
@@ -100,7 +99,8 @@ func NewValue(v interface{}) Value {
 	case float64:
 		return NewFloatValue(val)
 	case uint:
-		if !Buffer.Arch64Bits || (val <= math.MaxInt64) {
+		maxInt := ^uint(0) >> 1
+		if !Buffer.Arch64Bits || (val <= maxInt) {
 			return NewLongValue(int64(val))
 		}
 	case []interface{}:


sparrc:
to prove that this code works:

```
% cat /tmp/print.go
package main

import (
	"fmt"
)

func main() {
	fmt.Println(^uint(0) >> 1)
}
% GOARCH=386 go run /tmp/print.go
2147483647
% go run /tmp/print.go
9223372036854775807
```

khaf:
No Problem with the code, I will check and include it in the next release :)

Are you using the library in a 32bit platform? Can you expand a bit on your use case and environment?

sparrc:
yes, I am using the library in Telegraf: https://github.com/influxdata/telegraf

It is a plugin-based metrics collector for the InfluxData platform, we are using the client library for the aerospike plugin. The metrics can then be written into time-series platforms like influxdb, graphite, datadog, etc.

I'm not 100% sure if we have any users monitoring aerospike from 32-bit systems, but it at least helps our build process for this library to support 32-bit builds.

khaf:
In this case you should be very careful with the results, since the library will return the results as `int64` in 32bit platforms while it will return them as `int` in 64bit ones. That will affect you if you are doing direct `rec.Bins["bin"].(int)` type casts. Ideally, you will need a function to do it conditionally.

sparrc:
that should be OK, somewhere up the stack we convert every type of int to an int64, I'm not aware of any type switching beyond that.

Issue #75, opened 2015-07-26
Got many "EOF" errors after upgrade to 1.6.4


quexer:
- we got 6000+ "EOF" errors every day after upgrade to 1.6.3 
- when we rollback to 1.6.2, the number reduce to 100
- we use a 3-nodes aerospike cluster in LAN, and the LAN is high speed and very stable
- any advice ?

thanks!

GeertJohan:
Are you sure you are using 1.6.3, not 1.6.4?
Because 1.6.4 adds some error checking code that could display more EOF's. See commit 0f1e4cd.
They aren't occuring more often aince that commit, they're just handled better.

Do you have more information than just "EOF" ?

khaf:
What command causes the EOF errors? (Get, Put, Query, ...)

quexer:
@GeertJohan we check if the err  after every aerospike operation. if it isn't nil,  print the err.  then we go many "EOF" error in log file
@khaf  both "GET", "PUT", "QUERY" cause the EOF errors

quexer:
@GeertJohan sorry I made a mistake.
Yes, we're using 1.6.4 and it show more "EOF" errors than 1.6.3.

any update about this issue?

khaf:
This is expected if you are getting those EOF errors from queries. The old code was ignoring these errors, hence you weren't getting them.

Are these EOF errors impacting your operations? I would guess that increasing the timeout for queries would help a bit. 

Any guess regarding the percentage of bad requests to total?

quexer:
There EOF errors are impacting my operation, although the percentage is not very high ( about 0.01 %)

How can I fix this ? Should I add retry logic to my application ?  
Can "GET" operation get benefit from  increasing the timeout for queries   ?

thanks！

khaf:
I have an idea of what may be the problem. Let me work on it a bit and I'll come back to you.

Retries are baked in the client, and if you look at [`BasePolicy`](https://github.com/aerospike/aerospike-client-go/blob/master/policy.go), you can use `MaxRetries` and `SleepBetweenRetries` so that the client does the retries for you. `BasePolicy` is embedded in all policy objects.

quexer:
I've dug into the code. When we do any operation with nil policy, the default policy will be used. 
It looks the default value "MaxRetries=2" doesn't help

quexer:
@khaf any update ?

khaf:
[...]

PR #143, opened 2016-09-07
Retry in-case of EOF on parseResult

hectorj:
Related to #75 and https://github.com/aerospike/aerospike-server/issues/156

According to https://github.com/aerospike/aerospike-client-java/blob/master/client/src/com/aerospike/client/command/SyncCommand.java, all IO errors (even during parseResult) should lead to a retry.

Not very happy with the code duplication.
Improvements are welcome.
Sadly I don't have the time nor the right idea yet to refactor the error handling in this func.

### DIFF
--- a/command.go
+++ b/command.go
@@ -17,11 +17,12 @@ package aerospike
 import (
 	"errors"
 	"fmt"
+	"io"
 	"time"
 
 	. "github.com/aerospike/aerospike-client-go/logger"
-
 	. "github.com/aerospike/aerospike-client-go/types"
+
 	ParticleType "github.com/aerospike/aerospike-client-go/types/particle_type"
 	Buffer "github.com/aerospike/aerospike-client-go/utils/buffer"
 )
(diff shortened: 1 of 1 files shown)


edbo:
Is there any timeline on this being brought into the main client? We relatively often get EOF errors coming in and sometimes it tends to hit in batches for some reason.

hectorj:
@edbo : whether this or another similar PR gets merged, the best fix is to follow this recommendation: https://github.com/aerospike/aerospike-client-java/blob/66a0899462b89e85f8cca03e0ecd2d05d6035386/client/src/com/aerospike/client/policy/ClientPolicy.java#L65-L67

Set your client's [IdleTimeout](https://github.com/aerospike/aerospike-client-go/blob/9a5793109e5213056a459996ba0facf104186f71/client_policy.go#L39) lower than your server's `proto-fd-idle-ms`.

This PR will only make the client retry in case it does hit this error, but its better to not hit it at all.

edbo:
We've done all of that already. For some reason it seems to happen to around 50% of requests all at once and across multiple application servers. Then it miraculously stops and goes back to normal.

khaf:
Sorry for the delay in merging this PR and thanks for your patience. We'll get it into the client in the next release latest early next week.

khaf:
Merged and in master.

PR #152, opened 2016-11-06
PutObject: handle blobs correctly

blide:
Current PutObject implementation stores []byte as LIST aerospike type. That differs from non-reflection methods behavior


blide:
@khaf any reply?

khaf:
@blide Makes sense, will include it in the next release. Thanks for your pull request and patience.

khaf:
Thanks for your PR.

PR #153, opened 2016-11-07
recordset.go Results goroutines leak fix

deepakprabhakara:
The cancelled channel needs to be closed in signalEnd, otherwise we end up with leaked goroutines. This is because the goroutine started in Results() ends up waiting on the select. I use a select/default block to ensure the cancelled channel is closed only once since it is also closed in Close().

### DIFF
--- a/recordset.go
+++ b/recordset.go
@@ -161,7 +161,11 @@ L:
 //    }
 //  }
 func (rcs *Recordset) Results() <-chan *Result {
-	res := make(chan *Result, len(rcs.Records))
+	recCap := cap(rcs.Records)
+	if recCap < 1 {
+		recCap = 1
+	}
+	res := make(chan *Result, recCap)
 
 	select {
 	case <-rcs.cancelled:
@@ -202,8 +206,13 @@ func (rcs *Recordset) Close() error {
 		return ErrRecordsetClosed
 	}
 
+	// check and close cancelled channel
 	// this will broadcast to all commands listening to the channel
-	close(rcs.cancelled)
+	select {
+	case <-rcs.cancelled:
+	default:
+		close(rcs.cancelled)
+	}
 
 	// wait till all goroutines are done, and signalEnd is called by the scan command
 	rcs.wgGoroutines.Wait()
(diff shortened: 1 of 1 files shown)


deepakprabhakara:
tests failing, will resubmit after fixing them.

deepakprabhakara:
Found a better fix, tests should pass now.

recordset.go
```diff
@@ -197,14 +201,19 @@ func (rcs *Recordset) Results() <-chan *Result {
...
 // subsequent calls to the method will return ErrRecordsetClosed.
 func (rcs *Recordset) Close() error {
+	// check and close cancelled channel
+	// this will broadcast to all commands listening to the channel
+	select {
+	case <-rcs.cancelled:
+	default:
+		close(rcs.cancelled)
+	}
+
```
khaf: I don't see the benefit of this block. cancelled channel is already being closed exactly once. What am I missing?

ps. Sorry for the late reply, I had replied a few days ago and I just found out that when you begin a review, you should also explicitly finish it.

deepakprabhakara:
The issue is that in signalEnd() when the goroutines count reached 0 then rcs.active.Set gets called. This causes the close of the cancelled channel to get skipped when Close() is called (because it is after the rcs.active.CompareAndToggle(true) call)

deepakprabhakara:
So essentially in the current code cancelled channel gets closed once only if signalEnd() hasn't already finished. If signalEnd() finishes before Close() then cancelled never gets closed and goroutines start leaking.

recordset.go
deepakprabhakara: The issue is that in signalEnd() when the goroutines count reached 0 then rcs.active.Set gets called. This causes the close of the cancelled channel to get skipped when Close() is called (because it is after the rcs.active.CompareAndToggle(true) call).

So essentially in the current code cancelled channel gets closed once only if signalEnd() hasn't already finished. If signalEnd() finishes before Close() then cancelled never gets closed and goroutines start leaking.

The block I added ensures cancelled get closed (only once) but is not blocked by the state of active.

khaf:
This was a more complicated situation, which couldn't be fixed without a lock.

Thanks for finding and reporting it.

deepakprabhakara:
Excellent! You're most welcome. Thanks for the review and merge.

PR #156, opened 2016-11-23
Added chnages in putObject so that it supports objects instead of onl…

sarathsp06:
…y object pointers

Changes
1. Removed the assumption that every object send to putObject is a pointer
2. Removed some unnecessary else loops which follows  retrun statemnts
3. Replaced reflect.Indirect to obj.Elem as that is what essentially Indirect is doing and there is an extra Kind() check


khaf:
Thanks Sarath, will include this in the next release.

PR #168, opened 2016-12-18
In case of LimitConnectionsToQueueSize is false, connectionCount is invalid

bluele:
Hi there.
It seems that `connectionCount` is always invalid value when `LimitConnectionsToQueueSize` is false.

### DIFF
--- a/node.go
+++ b/node.go
@@ -442,8 +442,10 @@ func (nd *Node) getConnection(timeout time.Duration) (conn *Connection, err erro
 	}
 
 	if conn == nil {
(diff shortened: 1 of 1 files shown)


khaf:
In this case the better idea would be to reverse the order to avoid the short circuit, don't you think?

bluele:
Yes. In addition, I think that an explicit declaration would be better.

khaf:
An explicit declaration is not good, since two different goroutines can call it and move on to the next section which would deny both of them the connection. Just having it in one line avoids that problem.

PR #169, opened 2016-12-20
rename "internal" to "lib" so this library plays nicely with vendoring

Abramovic:
Projects that have vendoring enabled get this error: 

```bash
imports github.com/aerospike/aerospike-client-go/internal/lua: use of internal package not allowed
```

How to reproduce: 

1. Have a go project with vendoring enabled (using the vendor/ directory) 
2. Remove $GOPATH/src/github.com/aerospike/aerospike-client-go to make sure that we use the library inside of the vendor directory


khaf:
I think the issue you are facing is discussed here, and working around it does not require any changes to the library:

https://github.com/golang/oauth2/issues/178

Abramovic:
The proposed solution from that thread is to remove the vendor directory ``need rm -rf vendor/`` https://github.com/golang/oauth2/issues/178#issuecomment-260294124 but if a developer wants to use vendoring then this won't work. 

Here's a similar issue from the same repo that recommends doing a find and replace for "internal" using sed: https://github.com/golang/go/issues/12282#issuecomment-134154024

khaf:
From what I understand from that thread, what's happening is that the vendoring solution you are using does not copy the `internal/` sub-dir to your `vendor` dir and you may need to do it manually.

PR #172, opened 2016-12-30
Fix panic in GetObject() if all bins in results nil

Hamper:
For GetObject() for
```go
struct {
  Test int `as:"test"`
}
```
if bin 'test' not exists we get 'panic: reflect: call of reflect.Value.Type on zero Value'


khaf:
Merged, thanks.

PR #174, opened 2017-01-06
Fix WritePolicy usage with UDF

### LINKED ISSUE #173: UDF and TTLDontUpdate
TTLDontUpdate seems to not be honored when used with ``client.Execute``

* ``Touch`` a record with a TTL of 200
* Call an UDF function on it ``client.Execute``, with a write policy using Expiration set to ``TTLDontUpdate``
* Check the TTL of the record: seems to be set to namespace default TTL.

UDF used for test: this [one](https://github.com/bpaquet/aerodis/blob/22628e79fc595e279d9c2dabc0bfa9923a13d002/redis.lua#L163) 

With ``client.PutBins`` instead of ``client.Execute``, and the same write policy, the TTL is modified.

bpaquet:
Fix #173

### DIFF
--- a/command.go
+++ b/command.go
@@ -372,7 +372,7 @@ func (cmd *baseCommand) setUdf(policy *WritePolicy, key *Key, packageName string
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


khaf:
Merged. Thanks.

PR #182, opened 2017-03-17
Ignore hosts from other subnets

wedi-dev:
Not common usecase. service.address set to 0.0.0.0 and some interfaces not visible beacuse of proxy or VPN. Initialising client takes forever.


khaf:
Thanks for the pull request.

Did you have a look at clientPolicy.IpMap? Doesn't it cover the same general use-case?

We'll review the change and will merge upon approval for the next release.

wedi-dev:
That's right - IpMap indeed helped, but it's pretty hard to debug which adress client tries to connect, that's why we are doing filtering.

PR #190, opened 2017-04-11
minor: missing ResultCode types

alicebob:
All the other constants are of the type `ResultCode`, so these should be as well.


khaf:
Thanks for you PR. Merged :)

PR #191, opened 2017-04-21
Fix ClientPolicy timeout default value commentary

nassor:


khaf:
Thanks for you PR. Merged :)

PR #200, opened 2017-07-20
don't share the onComplete channel

alicebob:
Sharing the onComplete between multiple readers breaks the error return mechanism. And if it's an unusual use-case to have multiple readers (which I expect) there is no need to keep it around. It also simplifies the loop code a bit.

Second commit removes an unused variable.

The two commits are not related, really.

### DIFF
--- a/task.go
+++ b/task.go
@@ -30,14 +30,12 @@ type Task interface {
 type baseTask struct {
 	retries int
 	cluster *Cluster
-	done    bool
 }
 
 // newTask initializes task with fields needed to query server nodes.
-func newTask(cluster *Cluster, done bool) *baseTask {
+func newTask(cluster *Cluster) *baseTask {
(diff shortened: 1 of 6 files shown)


task.go
```diff
@@ -28,51 +28,39 @@ type Task interface {
...
-					btsk.onCompleteChan <- err
-					return
-				} else if done {
-					btsk.onCompleteChan <- nil
-					return
-				}
-			} // select
+			<-time.After(interval)
+			done, err := ifc.IsDone()
+			btsk.retries++
```
khaf: There is a race condition here since there can be multiple readers.

alicebob: But that was there in the orignal code as well. That didn't change.

khaf:
Couldn't merge this PR directly. Added it manually. Thanks.

PR #201, opened 2017-07-20
(tiny) functions never return an error

alicebob:
No need to pretend to return an error if you never do.


khaf:
These dummy errors are for future compatibility reasons. We may do validations on the filters in the future.

PR #202, opened 2017-08-02
GetConnectionCount metrics for node and client

komly:
Hi guys!
 We need to monitor connections count with nodes but we have no handle to lookup on it.
 Its simple commit to fetch its count via client.
 Could you please review my pull request.
 Thank you!

### DIFF
--- a/client.go
+++ b/client.go
@@ -130,6 +130,16 @@ func (clnt *Client) GetNodeNames() []string {
 	return names
 }
 
+// GetConnectionCount return sum connection count of all active nodes in cluster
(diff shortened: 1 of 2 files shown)


khaf:
Thanks for your pull request.

While I don't see a problem with this pull request, I'm reluctant to include it since I have bigger plans in mind regarding monitoring/instrumentation of the client.

At the moment I'm on vacation, I'll release a new client before August 20th with that API.

khaf:
I just released the new version. You can call the `Client.Stats()` and get a detailed count per node. We'll add more info there over time.

PR #224, opened 2018-02-22
Added latency metrics for aerospike

shubham340:
Added latency metrics for aerospike. 
Metrics include ops/sec ,  >1ms, >8ms, >64ms for read and write operation for each namespace.

### DIFF
--- a/info.go
+++ b/info.go
@@ -17,9 +17,9 @@ package aerospike
 import (
 	"bytes"
 	"encoding/binary"
+	"regexp"
 	"strings"
 	"time"
-
 	. "github.com/aerospike/aerospike-client-go/logger"
 	. "github.com/aerospike/aerospike-client-go/types"
 )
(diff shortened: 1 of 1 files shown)


khaf:
What is the rational for this to be inside the client code? There are other values like throughput that could also be added, but should they?

shubham340:
Yes, they should be added. Ideally, we should have a single source for all fetching kinds of metrics. It should depend upon the end user/customer which metrics to track and which metrics to ignore.

khaf:
The client code is already big and complicated enough. There's no reason one could not do this outside the client in a different codebase.

shubham340:
We are already fetching latency metrics for our application. We thought these metrics can be helpful for the entire aerospike community who track these metrics.

khaf:
Thank you for your contribution.

Let us discuss this internally and see if there is a consensus regarding bringing this second tier functionality inside the client code.

khaf:
After internal talks, we came to the conclusion that we would like to have utility code outside the client.

Thanks for your contribution.

Issue #235, opened 2018-06-20
App Engine Standard compatibility/Remove Gopher-Lua dependency


adhiravishankar:
This library should aim to have it as App Engine compatible as possible because it definitely restricts usage. The main issue is that App Engine Standard does not support the gopher-lua library and yet, it's being used in this client library. I'm wondering how feasible it would be to have the dependency removed? Or have an app-engine compatible version like some libraries do?

khaf:
It is easy to make an app engine compatible version using tags. The main question is how to keep it that way? Do you know a way of testing against app engine to ensure compatibility?

adhiravishankar:
Yeah, you can use app engine testing context rather than the regular context to ensure compatibility. App Engine has an aetest package that one can use to ensure compatibility with App Engine. 

I do not know of any issues App Engine has compatibility wise other than the fact that it uses a custom version of the httpclient, context, and no native code (which is where the Lua code here conflicts).  If you look at https://github.com/mailru/easyjson/search?utf8=%E2%9C%93&q=appengine&type=    the code is essentially separated into similar interfaces but one with app engine compatible code and one without. So, depending on the build tag, it just builds and runs. It doesn't run a test specifically for app engine. But, I don't know of any libraries (including Google's that has a test suite specifically for Google App Engine). I will have to look more into that.

khaf:
This issue is resolved in the just released v1.34.0

PR #236, opened 2018-06-21
separate lualib from rest of library for app engine

adhiravishankar:
#235 this should fix the lua stuff for app engine


khaf:
The v1.34.0 release allows you to take out this functionality via the `app_engine` build tag.

PR #259, opened 2019-04-15
Don't send racks: command if clientPolicy.RackAware is not set

kolo:
Hi,

this change updates `Node#Refresh` function to send `racks:` command based on `clientPolicy.RackAware` flag. There are two reasons behind it:

1. `racks:` won't be sent to older versions of `Aerospike` which don't support it. This will remove `received command racks, not registered` message from the log which is getting spammed by it otherwise.
2. `updateRackInfo` function [checks the state of the flag](https://github.com/aerospike/aerospike-client-go/blob/master/node.go#L232-L234) as well and doesn't use result of the command if it is not set.

### DIFF
--- a/node.go
+++ b/node.go
@@ -124,7 +124,12 @@ func (nd *Node) Refresh(peers *peers) error {
 	var infoMap map[string]string
 	var err error
 	if peers.usePeers.Get() {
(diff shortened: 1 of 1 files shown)


khaf:
Thanks for the PR. Will include it in the next version.

PR #263, opened 2019-08-13
Support write only (write) privilege for user roles

spkesan:


khaf:
This was already done internally.

Issue #272, opened 2019-12-10
Goroutine stack exceed on udf parameters unwrapping


small-egg:
When I send types like `as.MapValue` or `as.ListValue` to `QueryAggregate`, I get such error because `GetObject` method returns object without type conversion to the specific type. 
So there is infinity recursion [there](https://github.com/aerospike/aerospike-client-go/blob/master/value.go#L1135).

Would you like to fix it?

khaf:
Thanks for your PR. Merged in the latest release (v2.7.0)

PR #273, opened 2019-12-10
Issue #272 fix

small-egg:
https://github.com/aerospike/aerospike-client-go/issues/272 issue-related fix


khaf:
Thank you for this pull request. Could you please add tests in `client_object_test.go` for the cases you fixed?

small-egg:
Of course! I've done it in the last few commits. 
Sorry for so many commits. I couldn't run tests locally because there are imports of internal packages, and I get errors like `client.go:30:2: use of internal package github.com/aerospike/aerospike-client-go/internal/atomic not allowed`

PR #284, opened 2020-02-14
Simple example using predicate expressions with go client

a2chang:


khaf:
Thanks André, I'll include this in the next release.

Issue #180, opened 2017-02-07
Embedded structs in aerospike-go library unexpected behaviour


sarathsp06:
Here is the problem explained : http://stackoverflow.com/q/41931030/1898494

**Problem**

Need to add set of common fields to all the sets ,ie CreatedAt,UpdtedAt,DeletedAt etc. For the same I have created a struct and embed that with all the set structs. I need the Fields of the common structure saved in the set as fields of the given record

Sample Code

```go
type Table struct {
   CreatedAt time.Time
   UpdatedAt time.Time
   DeletedAt time.Time
}

type Account struct {
   Table
   Name string
   Status bool
   .....
}
```
For the above mentioned struct Account.I expect the record stored with bin names

```
CreatedAt,UpdatedAt,DeletedAt,Name,Status.....
```
But when the records are stored bin names are
```
Table,Name,Status...
```
Where Table would be a map with key values

Is it possible to achieve the expected behaviour ? if so how?

sarathsp06:
Any comments on this ??

khaf:
Needless to say it is not supported at this time, and I'm not sure the complexity involved to support this feature is worth the effort.

I'll keep this issue open fora while to see if there are others with similar use cases.

sarathsp06:
@khaf makes sense. It is actually the default behavior with `encode/json` and `gorm` .  So I was expecting it as the behavior in aerospike  case also .  I can contribute if it makes sense

khaf:
We are open to pull requests :)

PR #287, opened 2020-03-12
Add anonymous struct field support.

### LINKED ISSUE #286: PutObject does not respect embedded/anonymous fields
I create a test file in aerospike-client-go package

```go
package aerospike

import "testing"

type record struct {
	a
	b
}

type a struct {
	A int `as:"a"`
}

type b struct {
	B string `as:"b"`
}

func TestMarshal(t *testing.T) {
	var r record
	r.A = 10
	r.B = "b"
	v := marshal(r, true)
	t.Log(v)

	// expected
	t.Log(map[string]interface{}{
		"a": 10,
		"b": "b",
	})
}
```

`go test -v -run Marshal` outputs `map[]` while the expected output should be `map[a:10 b:b]`

 https://github.com/jmoiron/sqlx/blob/master/reflectx/reflect_test.go#L38-L94 The implementation by sqlx is very elegant.

andot:
Fixed #286 & #180

### DIFF
--- a/anonymous_fields_test.go
+++ b/anonymous_fields_test.go
@@ -32,19 +32,25 @@ var _ = Describe("Aerospike", func() {
 		var set = randString(50)
 
 		type anonymousStructA struct {
-			A int `as:"a"`
+			A   int    `as:"a"`
+			TTL uint32 `asm:"ttl"`
+			GEN uint32 `asm:"gen"`
 		}
 
 		type anonymousStructB struct {
-			B string `as:"b"`
+			B   string `as:"b"`
+			TTL uint32 `asm:"ttl"`
+			GEN uint32 `asm:"gen"`
 		}
 
 		type anonymousStructABC struct {
 			anonymousStructA
 			anonymousStructB
-			A bool    `as:"ace"`
-			B int     `as:"bce"`
-			C float64 `as:"c"`
+			A          bool    `as:"ace"`
+			B          int     `as:"bce"`
+			C          float64 `as:"c"`
+			TTL1, TTL2 uint32  `asm:"ttl"`
+			GEN1, GEN2 uint32  `asm:"gen"`
 		}
 
 		type anonymousStructABCD struct {
@@ -81,10 +87,9 @@ var _ = Describe("Aerospike", func() {
 				err = client.PutObject(nil, key, expected)
 				Expect(err).ToNot(HaveOccurred())
 
-				var actual testStruct
-				err = client.GetObject(nil, key, &actual)
+				actual := &testStruct{}
+				err = client.GetObject(nil, key, actual)
 				Expect(err).ToNot(HaveOccurred())
-				Expect(actual).To(Equal(*expected))
 
 				rec, err := client.Get(nil, key)
 				Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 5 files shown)


khaf:
@andot Thanks for the PR. Could you please add a good test in the `client_reflect_test.go` to make sure things work?

andot:
@khaf I add a new test file `anonymous_fields_test.go`.

khaf:
@andot Thanks. I believe the mapping code is not sophisticated enough and will fail. Take the following example:
```go
var _ = Describe("Aerospike", func() {

	Describe("PutObject and GetObject with anonymous fields", func() {
		// connection data
		var err error
		var ns = *namespace
		var set = randString(50)

		type anonymousStructA struct {
			A int `as:"a"`
		}

		type anonymousStructB struct {
			B string `as:"b"`
		}

		type anonymousStructABC struct {
			anonymousStructA
			anonymousStructB
			A bool    `as:"ace"`
			B int     `as:"b"`
			C float64 `as:"c"`
		}

		type anonymousStructABCD struct {
			ABC *anonymousStructABC
			D   bool `as:"d"`
		}

		type testStruct struct {
			anonymousStructABC
			anonymousStructABCD
		}

		makeTestObject := func() *testStruct {
			obj := &testStruct{}
			obj.A = true
			obj.B = 10
			obj.anonymousStructA.A = 10
			obj.C = 3.14159
			obj.ABC = &anonymousStructABC{}
			obj.ABC.A = false
			obj.ABC.B = 42
			obj.ABC.anonymousStructB.B = "World!"
			obj.ABC.C = 2.17828
			obj.D = true
			return obj
		}

		Context("PutObject & GetObject operations", func() {
			It("must save an object with anonymous fields", func() {
				key, _ := as.NewKey(ns, set, randString(50))
				expected := makeTestObject()
				err = client.PutObject(nil, key, expected)
				Expect(err).ToNot(HaveOccurred())

				var actual testStruct
				err = client.GetObject(nil, key, &actual)
				Expect(err).ToNot(HaveOccurred())
				Expect(actual).To(Equal(*expected))
			})
		})
	})
})
```
Since there are multiple `B` fields, the unmarshal has trouble finding out which value belongs to which, and panics.
The mapper code should be enhanced to support this use case when the types are incompatible (like `int` and `string` for B), but I worry about the compatible types (like `bool` and `int` or even `int` and `int` for A).
So I guess the mapping should become a `map[string]interface{}` to allow for nested maps.

andot:
I think change mapping from `map[string]string` to `map[string][]int` is better than `map[string]interface{}`, the `[]int` element stores the Index of the field.

```go
		type anonymousStructB struct {
			B string `as:"b"`
		}

		type anonymousStructABC struct {
			anonymousStructA
			anonymousStructB
			A bool    `as:"ace"`
			B int     `as:"b"`
			C float64 `as:"c"`
		}
```

But I think the `B` in `anonymousStructB` and `anonymousStructABC` is conflicted by tag. Define different tags can solve this problem.

khaf:
I believe the mapper can easily detect this conflict and panic if it detects the conflict exists.

andot:
@khaf The new test is passed.

andot:
I will change the metaMappings (ttlMap, genMap) from `[]string` to `[][]int` too.

khaf:
Can you please also make sure that ambiguous fields with the same name or alias are not allowed and the library panics if it detects such a situation?

andot:
@khaf I find in https://github.com/aerospike/aerospike-client-go/blob/5990c8c084f2e2d6da98598312b333a29d661bda/client_object_test.go#L307-L309

`EmpNstdObjSlic` and `EmpNstdObjPSlc` have the same alias, the modified code detects it and panics.

khaf:
perfect!

khaf:
I don't understand how the test for `TTL` fields work. I expected that the `TTL` fields for `anonymousStructA` and `anonymousStructB` to conflict with each other and panic, since they have the same level of precedence and have the same tag. Do you treat the meta tags differently than the data tags?

andot:
https://github.com/aerospike/aerospike-client-go/blob/5990c8c084f2e2d6da98598312b333a29d661bda/client_object_test.go#L822-L823

TTL1, TTL2 is also in the same level of precedence and have the same tag, but the test didn't treat them  to conflict with each other and panic. so I did the same thing.

khaf:
Fair enough. They will be the same value anyway.

andot:
@khaf  Can you merge this PR? We need this feature urgently in production.

khaf:
Released in v2.9.0. Thank you for your contribution.

PR #292, opened 2020-03-26
Add option that custom dial function for creating tcp connections

kazukousen:
this PR is adds support for customizable dial function.

e.g. over SSH
```go
sshClient, err := ssh.Dial("tcp", fmt.Sprintf("%s:%d", host, 22), sshConfig)
policy := aerospike.NewClientPolicy()
policy.Dial = sshClient.Dial
```

### DIFF
--- a/client_policy.go
+++ b/client_policy.go
@@ -16,6 +16,7 @@ package aerospike
 
 import (
 	"crypto/tls"
+	"net"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


khaf:
May I ask why this is needed while the server itself supports TLS connections?

kazukousen:
thank you for your reply.
It is not the purpose of the server itself to support TLS connections.
it is for performing tasks such as multi-stage SSH connection from a laptop.

PR #297, opened 2020-05-06
fix: distinct error when empty slice is passed to BatchGetObjects

oss92:
- [x] separate error when `keys` slice is empty
- [x] fixed error as per go guidelines (no capitalization and don't end with punctuation)


khaf:
Thanks for your PR.

PR #302, opened 2020-06-17
GetObject slices bug: assign result of the append

gdm85:
There is currently a bug happening when using `GetObject()`. See the following pseudo-code:

```go
type MyObject struct {
    List []string
}

func testcase() {
    var o1, o2 MyObject
    o1.List = nil // the default is nil anyways
    o2.List = []string{}

    client.GetObject(nil, key, &o1)
    client.GetObject(nil, key, &o2)

    if len(o1.List) != len(o2.List) {
         panic("BUG: slice was not set properly")
    }
}
```

With the change proposed here the bug is fixed (the slice gets assigned); the assignment `f = ` there is currently doing nothing.

There do not seem to be unit tests for this function, otherwise I would have added a regression test; perhaps I can contribute something on a separate PR.


khaf:
Thanks for your PR. Merged in the latest release.

PR #314, opened 2020-08-31
fix policy document

sinozu:
Hi,

This fix updates policy document to match the code.
- Removed `GenerationPolicy.DUPLICATE`
- Update `Expiration` Values


khaf:
Thank you for the PR, will include it in the next release.

PR #328, opened 2020-12-03
Added latency metrics to Query and PutBins methods -- WIP

KolbyMcGarrah:
Initial changes for adding latency metrics to the aerospike go client Query and PutBins methods. 
I needed to change the command interface Execute, execute and executeAt functions to all include a context as a parameter so that I could propagate tags down to be recorded. This required updating all of the other instances of those functions which is why so many files were touched. 

Some key features that need to be implemented are: 
- Users will need to register the default views in order to start collecting metrics. 
- Users will need to utilize the client.SetInstance(name string) method to be able to set the cluster instance in order to include the instance name in the metrics. 
- Users will need to pass context in all aerospike calls when using this library even if the call isn't being collected in metrics.


khaf:
Sorry for the late reply, I was on Vacation for most of December. I'm curious as to why this PR was posted, and why it was subsequently closed immediately?

PR #335, opened 2021-01-18
Add String function to ResultCode

davidae:
Add a `String` function to `ResultCode` to have an
understandable/readable and constant string for all result codes.
This will f.ex. make monitoring easier, using `ResultCode.String()` as 
a label value on a counter without the need to lookup the code 
(integer) definitions.


khaf:
Sorry for the late response. I'll be reworking the whole Error system on the Go client for the next major release, so I'll leave this out of v4.x releases until I've had time to give it a bit of thought.

khaf:
This was done in v5. Thank you for your idea and feedback.

PR #340, opened 2021-02-14
Fix documentation

reugn:


khaf:
This was addressed in v4.3.0 release, and merging it would cause conflicts. Thanks for the PR though.

PR #381, opened 2022-06-29
Fix BatchOperate: use offset index to pick correct record

TvoroG:
BatchOperate sends same records to each node


TvoroG:
Thanks! Will close it

PR #407, opened 2023-06-23
Add a param in clientPolicy for using only seedNodes for connection

sud82:


khaf:
Could you add a bit of info regarding why this feature is needed?

sud82:
We are using this go client in our https://github.com/aerospike/aerospike-kubernetes-operator (AKO). AKO uses the server pod's IP as the seed address to connect to the server pod. The problem here is that if this podIP is different from the IP returned by the pod using the command `service-clear-std` then go-client thinks that podIP is the IP of a load balancer. The client will then use the IP fetched using the command `service-clear-std`. If this fetched IP is not exposed within the k8s cluster then AKO will not be able to connect to the server pod using this fetched IP.
This PR adds a field in the clientPolicy that ensures that the client uses only the seed IPs.

PR #430, opened 2024-04-23
chore: deprecate io/ioutil

swarit-pandey:
*Description*: Deprecate the `io/ioutil` pkg

*Breaking*: No

*Scope*: Only at places where `io/iotuil` was being used

*Tested*: Yes


khaf:
Thanks for your pull request. merged and released in v7.3.0.

PR #432, opened 2024-05-06
CLIENT-2910 Add Go Client feature request template

justinlee-aerospike:


khaf (approved):

khaf (commented):
Will this work if only merged in the v7 branch? I would expect that it would need to be merged in the main branch.

PR #436, opened 2024-05-23
Fix error handling in authInterceptor.login

reugn:


khaf:
@reugn Thanks for the PR. I think shadowing the original `err` [Here](https://github.com/aerospike/aerospike-client-go/pull/436/files#diff-1b9ee7e5d0a12bb17d45d0e8c6088d14d7b1caa3512bc3cedff512909e89f860L141) would be a better idea to prevent this issue to return in the future.

proxy_auth_interceptor.go
```diff
@@ -139,14 +139,14 @@ func (interceptor *authInterceptor) login() Error {
 
 	claims := strings.Split(res.GetToken(), ".")
 	decClaims, gerr := base64.RawURLEncoding.DecodeString(claims[1])
```
khaf, line 141: Let's remove the source of the issue here and replce all the `gerr` instances with `err`.

khaf (changes requested):

proxy_auth_interceptor.go
reugn: @khaf, the assignment of the `Error` type to `err` and the `error` type to `gerr` may have required an additional error variable.

khaf: I see; thanks for the update.

PR #441, opened 2024-07-08
Resolve memory leak caused by circular reference in client finalizer

reugn:
A [circular reference](https://github.com/aerospike/aerospike-client-go/blob/v7.5.0/client.go#L115) between the client object and the cluster object prevents the [client's finalizer](https://github.com/aerospike/aerospike-client-go/blob/v7.5.0/client.go#L117) from executing, resulting in a memory leak. The issue can be reproduced by creating and closing multiple client instances in a loop.
To resolve this, the client instance has been injected into the batch commands directly.

### DIFF
--- a/batch_command.go
+++ b/batch_command.go
@@ -28,12 +28,18 @@ type batcher interface {
 	generateBatchNodes(*Cluster) ([]*batchNode, Error)
 	setSequence(int, int)
 
-	executeSingle(*Client) Error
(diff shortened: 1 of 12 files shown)


khaf:
@reugn Thanks for the pull request. Is the issue the finalizer not being called, or the memory leak after the finalizer call?

reugn:
The finalizer is not being called, causing the memory leak.

PR #443, opened 2024-07-17
Fix pointer value assignment in baseMultiCommand.parseKey

reugn:


khaf:
@reugn What is the issue this patch is trying to address?

reugn:
[This way](https://github.com/aerospike/aerospike-client-go/pull/443/files#diff-455db08555aac03f078b4d3f363a54b26eca2c1ff2f0cabdbfc1264312481656L225), the value assigned to `bval` is not visible outside of the method.

khaf:
Absolutely true, thanks.

PR #450, opened 2024-08-27
Add reference to client.WarmUp when an error about the connection pool shows up

bsergean:


khaf:
Rationale added to the original issue.

bsergean:
Agreed for a major version, but I would consider keeping a note to change it in a major one because this error is tricky and people don't ready read the documentation ...

Issue #446, opened 2024-08-15
Client v7 occasionally crushes with concurrent map read and map write


Gaudeamus:
Hello, seems like across different client versions the same bug is persistent
Current report is for client version v7.6.0

```
fatal error: concurrent map read and map write

goroutine 1314290423 [running]:
github.com/aerospike/aerospike-client-go/v7.PartitionForRead(0xc01b4086c0?, 0xc00693d860, 0xc3d4cebec0)
        /src/dsp/vendor/github.com/aerospike/aerospike-client-go/v7/partition.go:74 +0x95
github.com/aerospike/aerospike-client-go/v7.newReadCommand(0xc58bb36008, 0xc00693d860, 0xc3d4cebec0, {0xc3d8445860?, 0x5, 0x5}, 0x40?)
        /src/dsp/vendor/github.com/aerospike/aerospike-client-go/v7/read_command.go:54 +0x92
github.com/aerospike/aerospike-client-go/v7.(*Client).Get(0xc58bbb1f20, 0xc3d8445860?, 0xc3d4cebec0, {0xc3d8445860, 0x5, 0x5})
        /src/dsp/vendor/github.com/aerospike/aerospike-client-go/v7/client.go:475 +0x131
gitlab.mgid.com/services/dsp/aerospike.(*Client).Get(0x1739518?, 0xc?, 0xc2c649ea00?, {0xc3d8445860?, 0x1426740?, 0xc6c676c660?})
        /src/dsp/aerospike/client.go:126 +0x91
gitlab.mgid.com/services/dsp/aerospike.(*Session).GetDataByClientUserId(0xc08ee9ec30, {0xc2c649ea00, 0x6}, {0xc32db8c018?, 0x1?})
....
```

Note: same issue was reported for v6,v5 https://github.com/aerospike/aerospike-client-go/issues/399

khaf:
Thanks for your report. I'm working on it.

ravisastryk:
I understand everyone has been busy, but I made an attempt to fix this issue. Please take a moment to review it when you can so that everyone can benefit.

Here is my diff for your review https://github.com/aerospike/aerospike-client-go/pull/454

cc @khaf @Gaudeamus

khaf:
> Hello, seems like across different client versions the same bug is persistent Current report is for client version v7.6.0
> 
> ```
> fatal error: concurrent map read and map write
> 
> goroutine 1314290423 [running]:
[...]

Issue #399, opened 2023-03-26
Client v6 occasionally crushes with concurrent map read and map write


Gaudeamus:
client version v6.12.0
working with remote cluster on slow network I'm facing with a regular problem causing entire app to crush:

```
fatal error: concurrent map read and map write

goroutine 87917046 [running]:
github.com/aerospike/aerospike-client-go/v6.PartitionForRead(0xc0123d3cb8?, 0xc0999c7a90, 0xc059c39680)
        /src/vendor/github.com/aerospike/aerospike-client-go/v6/partition.go:66 +0xa5
github.com/aerospike/aerospike-client-go/v6.newReadCommand(0xc317e6f200, 0xc0999c7a90, 0xc059c39680, {0xc25e780d20?, 0x5, 0x5}, 0x10?)
        /src/vendor/github.com/aerospike/aerospike-client-go/v6/read_command.go:53 +0x90
github.com/aerospike/aerospike-client-go/v6.(*Client).Get(0xc318983da0, 0x50?, 0xc012580000?, {0xc25e780d20, 0x5, 0x5})
        /src/vendor/github.com/aerospike/aerospike-client-go/v6/client.go:355 +0x137

```

the same issue was with v5
```
github.com/aerospike/aerospike-client-go/v5.PartitionForWrite(0x0?, 0xc0a44b3830, 0xc4d432d800)
        /src/dsp/vendor/github.com/aerospike/aerospike-client-go/v5/partition.go:52 +0x96
github.com/aerospike/aerospike-client-go/v5.newExecuteCommand(0x7?, 0xc0a44b3830, 0x3?, {0x146b1ff, _}, {_, _}, _)
        /src/dsp/vendor/github.com/aerospike/aerospike-client-go/v5/execute_command.go:35 +0x9a
github.com/aerospike/aerospike-client-go/v5.(*Client).Execute(0xc3ec36ab40, 0xc01bc20970?, 0xc3bc5f?, {0x146b1ff, 0x9}, {0x146dec9, 0xb}, {0xc4d69cecb0, 0x1, 0x1})
        /src/dsp/vendor/github.com/aerospike/aerospike-client-go/v5/client.go:738 +0x25c
```

khaf:
Thanks for your feedback, I've been working on it since yesterday. Can you provide a bit more information regarding your cluster? How many nodes do you have and what usually prompts the issue (nodes departing the cluster, etc.)

khaf:
[...]

PR #454, opened 2024-10-20
Fix concurrent map read/write in partition management

ravisastryk:
**Problem**
The current implementation of `partitionMap` is not thread-safe, leading to potential race conditions during concurrent read and write operations. This issue affects various parts of the codebase, including cluster management, node operations, and partition parsing.

**Issue References**
https://github.com/aerospike/aerospike-client-go/issues/446
https://github.com/aerospike/aerospike-client-go/issues/399

**Solution**
Implement a thread-safe `partitionMap` using `sync.Map` and update all related functions to use the new thread-safe methods. This change ensures safe concurrent access to partition data across multiple goroutines.

**Technical Notes**
`partitionMap` is a thread-safe map that stores partition information for different namespaces. It uses a `sync.Map` internally to provide concurrent read/write access without explicit locking. The keys are namespace names (strings), and the values are pointers to Partitions structs. This structure allows for efficient, concurrent operations on partition data across multiple goroutines.

```
Usage:
 - Use get(key) to retrieve partition data for a namespace
 - Use set(key, value) to update or add partition data for a namespace
 - Use iterate(func) to iterate over all namespace-partition pairs safely
 - Use delete(key) to remove partition data for a namespace
 - Use len() to get the number of namespaces in the map
```

**Changes**
partitions.go

* Redefined `partitionMap` to wrap a `sync.Map`
* Implemented thread-safe methods: get, set, delete, len, and iterate
* Updated clone and cleanup methods to work with the new structure

partition.go

* Modified `PartitionForWrite` and `PartitionForRead` to use new get method of `partitionMap`

partition_parser.go

* Updated partitionParser struct to use a pointer to `partitionMap`
* Modified parsing logic to use new thread-safe methods of `partitionMap`

node.go

* Updated `refreshPartitions` method to work with the new `partitionMap` structure

cluster.go

* Changed `partitionWriteMap` in Cluster struct to use a pointer to `partitionMap`
* Updated `setPartitions`, `getPartitions`, and `findNodeInPartitionMap` methods
* Modified `NewCluster` function to initialize `partitionWriteMap` correctly
* Updated `tend` method to use thread-safe operations on `partitionMap`

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -55,7 +55,7 @@ type Cluster struct {
 	metricsPolicy  iatomic.TypedVal[*MetricsPolicy]
 
 	// Hints for best node for a partition
-	partitionWriteMap iatomic.TypedVal[partitionMap] //partitionMap
+	partitionWriteMap iatomic.TypedVal[*partitionMap] //partitionMap
 
 	clientPolicy        ClientPolicy
 	infoPolicy          InfoPolicy
@@ -130,7 +130,8 @@ func NewCluster(policy *ClientPolicy, hosts []*Host) (*Cluster, Error) {
 		supportsPartitionQuery: *iatomic.NewBool(false),
 	}
 
-	newCluster.partitionWriteMap.Set(make(partitionMap))
+	// Initialize partitionWriteMap with a new partitionMap
+	newCluster.partitionWriteMap.Set(newPartitionMap())
 
 	// setup auth info for cluster
 	if policy.RequiresAuthentication() {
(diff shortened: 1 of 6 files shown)


ravisastryk:
@khaf 
Can you please review this as soon as possible? It seems that many people are awaiting this fix. Let's aim to make it available to all.

cc @Gaudeamus

khaf:
@ravisastryk The `partitionMap` structure is a lockless construct. We have specifically avoided synchronizing the map. Using `sync.Map` has major performance consequences. As such, this PR is not workable.

PR #456, opened 2024-11-01
Go Client v7.7.2

### LINKED ISSUE #455: Client doesn't read from preferred rack
I was testing rack-aware reads but the results were always coming from the master node.

Traced this down to here:
https://github.com/aerospike/aerospike-client-go/blob/63370c8927591ab5d4a30e031c6046632be9ea3e/partition.go#L255C4-L255C9

`index` never changes inside the loop, which seems like a bug. changing the line to

```go
index := seq % len(replicas)
```

seems to fix it, but i'm not too familiar with the code so not going to make a PR.

khaf:
- **Fixes**
  - [CLIENT-3156] Fix an issue where rack policy always returns the master node. Resolves #455

[CLIENT-3156]: https://aerospike.atlassian.net/browse/CLIENT-3156?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Change History
 
+## November 1 2024: v7.7.2
+
+  Minor fix release.
+
+- **Fixes**
+  - [CLIENT-3156] Fix an issue where rack policy always returns the master node. Resolves #455
+
 ## September 18 2024: v7.7.1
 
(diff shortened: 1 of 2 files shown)


justinlee-aerospike (approved):
change looks good

PR #457, opened 2024-11-27
Parse nil keys properly in query operations

reugn:
Nil keys are valid and occur when writing a record using the `NewKeyWithDigest` constructor and the user key is nil. The client has to parse those records properly during a scan/query operation.

### DIFF
--- a/value.go
+++ b/value.go
@@ -1275,6 +1275,9 @@ func bytesToKeyValue(pType int, buf []byte, offset int, length int) (Value, Erro
 		}
 		return ListValue(v), nil
 
+	case ParticleType.NULL:
+		return NewNullValue(), nil
+
 	default:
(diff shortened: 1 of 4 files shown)


value.go
```diff
@@ -1275,6 +1275,9 @@ func bytesToKeyValue(pType int, buf []byte, offset int, length int) (Value, Erro
 		}
 		return ListValue(v), nil
 
+	case ParticleType.NULL:
+		return NewNullValue(), nil
```
khaf, line 1279: I believe this should return `nil, nil`.
`NullValue` implies the value was returned from the server, which is not possible.

khaf (changes requested):
Please make the change and test. I will merge as soon as possible.

value.go
reugn: Updated.

khaf (approved):

PR #458, opened 2024-12-11
Go Client v8.0.0-beta.1

khaf:
Major breaking release. This release supports Multi-Record Transactions. Please note that this is a beta release and will be subject to breaking changes.

- **Breaking Changes**
  - This release removes the proxy client and associated mock interfaces from the codebase. Maintaining those interfaces was very hard due to multiple build flags which had their own set of of supported API. Since interfaces are implicit, doing that by the user is as easy, or potentially easier outside of the library.
  - Operation used to return the result of multiple ops on a single bin wrapped in `[]any`. That would be confusing, since it was not clear if the result was the contents of the bin, or the results of multiple commands. In version 8, the client will wrap those results inside a `as.OpResults` instead of a `[]any`, removing the confusion.
  - The client would automatically populate the `stackFrame` field for the `as.AerospikeError`. Since that was an expensive operation, the client now will consult the value of `StackTracesEnabled` variable before doing so. The default is set to `false` of course.
  - The client now requires Go v1.23+. Needed to implement iterators later.

- **New Features**
  - [CLIENT-3159] Support writing raw payload to the server.
  - [CLIENT-3183] [CLIENT-3203] [CLIENT-3193] [CLIENT-3190] Support Multi-Record Transactions.

- **Fixes**
  - [CLIENT-3214] Remove Proxy code.
  - [CLIENT-3217] Do not send `nil` or `NullValue` as key to the server.

- **Improvements**
  - [CLIENT-3214] Support `-use-services-alternate` in test arguments.
  - Renamed all `interface{}` to `any`.

[CLIENT-3159]: https://aerospike.atlassian.net/browse/CLIENT-3159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


justinlee-aerospike (changes requested):
Changes look good.  Only issue is that we agreed this would be version 8.0.0 (beta tag okay).

justinlee-aerospike (dismissed):

justinlee-aerospike (approved):
changes look good

PR #465, opened 2025-02-19
Return error with the result code in parseResult

reugn:


khaf (changes requested):
Looks good. Could you please change the base to the stage branch? We usually merge everything there, test and then release into the main.

reugn:
Changed the base branch to `stage`.

khaf (approved):

ajkimdev:
```
func TestMrtBlocking(t *testing.T) {
	cp := aero.NewClientPolicy()
	cp.UseServicesAlternate = true
	cli, _ := aero.NewClientWithPolicyAndHost(cp, aero.NewHost("localhost", 3100))
	cli.WarmUp(-1)
	cli.Truncate(nil, "test", "", nil)

	count := 1

	var keys []*aero.Key
	for i := 0; i < count; i++ {
		key, _ := aero.NewKey("test", "demo", i)
		cli.PutBins(nil, key, aero.NewBin("bin", 1000))
		keys = append(keys,
			key,
		)
	}

	txn := aero.NewTxn()

	for _, key := range keys {
		p := aero.NewPolicy()
		p.Txn = txn
		if _, err := cli.Get(p, key); err != nil {
			fmt.Println(err)
		}
	}

	key0, _ := aero.NewKey("test", "demo", 0)
	cli.PutBins(nil, key0, aero.NewBin("bin", 999))

	status, err := cli.Commit(txn)
	if err != nil {
		fmt.Println(err)
		fmt.Println(err.Matches(-17))
	} else {
		fmt.Println(status)
	}

	cli.Close()
}
```
This is a test code to verify that MRT_VERSION_MISMATCH is always returned regardless of the number of records (the `count` var ). Please make sure to include this in your regression test plan.

PR #467, opened 2025-02-26
Fix nil check for error in executeIter method

lesha888:
Method executeAt can return nil


khaf (approved):

khaf:
Released in v8.1.0 and 7.9.0. Thanks.

PR #468, opened 2025-02-26
3347 - Returning error if resultCode is not OK

mirzakaracic:
Fixing reported bug where single record transaction conflicts were ignored.


khaf:
This was fixed via PR #465

PR #469, opened 2025-02-26
Check AerospikeError receiver for nil in exported methods

reugn:


mirzakaracic (approved):
The changes look good.

PR #473, opened 2025-03-05
Fix preferred rack on v6 client

mgebundy:
Cherry-picked the bug fix from https://github.com/aerospike/aerospike-client-go/pull/456 reported in https://github.com/aerospike/aerospike-client-go/issues/455


mgebundy:
fixed in https://github.com/aerospike/aerospike-client-go/pull/476 thank you!

PR #479, opened 2025-04-04
CLIENT-3365 Dynamic config changes

mirzakaracic:
Dynamic configuration changes for GO-lang client.


batch_delete_policy.go
```diff
@@ -58,7 +58,15 @@ func NewBatchDeletePolicy() *BatchDeletePolicy {
 	}
 }
 
-func (bdp *BatchDeletePolicy) toWritePolicy(bp *BatchPolicy) *WritePolicy {
+func NewBatchDeletePolicyOrDefaultFromCache(dynConfig *DynConfig) *BatchDeletePolicy {
```
khaf: I think a better name for it would be `NewDynamicBatchDeletePolicy`; Same for the rest of them.
The current name is not very Go like.

batch_delete_policy.go
```diff
@@ -71,5 +79,74 @@ func (bdp *BatchDeletePolicy) toWritePolicy(bp *BatchPolicy) *WritePolicy {
...
+		if config.Dynamic.BatchDelete.SendKey != nil {
+			wp.SendKey = *config.Dynamic.BatchDelete.SendKey
+		}
+	}
+
 	return wp
 }
+
+// copyBatchDeletePolicy creates a new BasePolicy instance and copies the values from the source BatchDeletePolicy.
+func copyBatchDeletePolicy(src *BatchDeletePolicy) *BatchDeletePolicy {
```
khaf: This should be a method:
```go
func (bdp *BatchDeletePolicy) copy() *BatchDeletePolicy {
...
}
```
Same for the rest.

batch_delete_policy.go
```diff
@@ -71,5 +79,74 @@ func (bdp *BatchDeletePolicy) toWritePolicy(bp *BatchPolicy) *WritePolicy {
...
+	if src == nil {
+		return nil
+	}
+
+	response := *src
+	return &response
+}
+
+// applyConfigToBatchDeletePolicy applies the dynamic configuration and generates a new policy
+func applyConfigToBatchDeletePolicy(policy *BatchDeletePolicy, dynConfig *DynConfig) *BatchDeletePolicy {
```
khaf: This also should be a method with a more idiomatic name:
```go
func (bdp *BatchDeletePolicy) patchDynamic(dynConfig *DynConfig) *BatchDeletePolicy {
...
}
```

batch_delete_policy.go
```diff
@@ -71,5 +79,74 @@ func (bdp *BatchDeletePolicy) toWritePolicy(bp *BatchPolicy) *WritePolicy {
...
+		responsePolicy = copyBatchDeletePolicy(policy)
+		responsePolicy = mapDynamicBatchDeletePolicy(responsePolicy, dynConfig)
+
+		return responsePolicy
+	} else {
+		return policy
+	}
+}
+
+func mapDynamicBatchDeletePolicy(policy *BatchDeletePolicy, dynConfig *DynConfig) *BatchDeletePolicy {
```
khaf: Same here:
```go
func (bdp *BatchDeletePolicy) mapDynamic(dynConfig *DynConfig) *BatchDeletePolicy {
...
}
```

bench_apply_configuration_test.go
```diff
@@ -0,0 +1,75 @@
...
+	for i := 0; i < b.N; i++ {
+		_ = applyConfigToClientPolicy(cp, dynCfg)
+	}
+}
+
+func BenchmarkApplyConfigToClientPolicyWithDynamicAndStaticConfig(b *testing.B) {
+	cp := NewClientPolicy()
+
+	cfg := &dynconfig.Config{
+		Static: &dynconfig.StaticConfig{
```
khaf, line 46: What were the result of discussions regarding removing the static/dynamic sections for future compatibility?

config.go
```diff
@@ -0,0 +1,499 @@
...
+
+	policy = mapDynamicMetricsPolicy(policy, dc)
+
+	return policy
+}
+
+// ----------------------------------------------------------------
+// Main watch goroutine for the config provider
+// ----------------------------------------------------------------
+func (dc *DynConfig) watchConfig(interval time.Duration) {
```
khaf, line 445: It's better for this interval to come from the config. Why does it not?

config/dynconfig.go
```diff
@@ -0,0 +1,329 @@
...
+	Enable         *bool `yaml:"enable"`
+	LatencyColumns *int  `yaml:"latency_columns"`
+	LatencyBase    *int  `yaml:"latency_base"`
+}
+
+// ----------------------------------------------------------------
+// Enum types
+// ----------------------------------------------------------------
+
+// TODO(Khosrow): Deal with the circular dependencies to remove the redefinition of these types.
```
khaf, line 199: Do the maps run just once when the config file is being read, or every time we patch user policies?

config/registry/registry.go
```diff
@@ -0,0 +1,55 @@
...
+package configregistry
+
+import (
+	"sync"
+
+	dynconfig "github.com/aerospike/aerospike-client-go/v8/config"
+)
+
+const (
+	DSN_REGEX_PATTERN = `^\s*([A-Za-z][A-Za-z0-9+.-]*://)?(.*)$`
```
khaf: It's best to use named patterns so that you won't have to fiddle with the indices.

txn_roll_policy.go
```diff
@@ -35,3 +37,79 @@ func NewTxnRollPolicy() *TxnRollPolicy {
...
+		// Passed in policy is nil, fetch mapped default policy from cache.
+		return dynConfig.client.dynDefaultTxnRollPolicy.Load()
+	} else if config != nil && config.Dynamic != nil && config.Dynamic.TxnRoll != nil {
+		// Dynamic configuration is exists for policy in question.
+		var responseTxnRollPolicy *TxnRollPolicy
+		// User has provided a custom policy. We need to apply the dynamic configuration.
+		responseTxnRollPolicy = copyTxnRollPolicy(policy)
+		responseTxnRollPolicy = mapDynamicTxnRollPolicy(responseTxnRollPolicy, dynConfig)
+
+		return responseTxnRollPolicy
```
khaf, line 69: By using methods, this would become a one liner:
```go
return policy.copy().mapDynamic(dynConfig)
```

khaf (changes requested):
The file `ginkgo.report` should also be removed from the repo and added to the `.gitignore` file.

config/dynconfig.go
```diff
@@ -0,0 +1,329 @@
...
+	// static config
+	ConfigInterval        *int `yaml:"config_interval"`
+	ConnectionQueueSize   *int `yaml:"connection_queue_size"`
+	MinConnectionsPerNode *int `yaml:"min_connections_per_node"`
+
+	// dynamic config
+	IdleTimeout         *int   `yaml:"max_socket_idle"`
+	Timeout             *int   `yaml:"timeout"`
+	ErrorRateWindow     *int   `yaml:"error_rate_window"`
+	MaxErrorRate        *int   `yaml:"max_error_rate"`
```
khaf, line 72: This and a few more values are set in the `ClientPolicy`, but the `clientPolicy` attribute in the Cluster object is not turned in to an `atomic.Pointer`. That change needs to be done.

config/dynconfig.go
```diff
@@ -0,0 +1,329 @@
...
+	BatchRead   *BatchRead   `yaml:"batch_read"`
+	BatchWrite  *BatchWrite  `yaml:"batch_write"`
+	BatchUdf    *BatchUdf    `yaml:"batch_udf"`
+	BatchDelete *BatchDelete `yaml:"batch_delete"`
+	TxnRoll     *TxnRoll     `yaml:"txn_roll"`
+	TxnVerify   *TxnVerify   `yaml:"txn_verify"`
+	Metrics     *Metrics     `yaml:"metrics"`
+}
+
+type Client struct {
```
khaf, line 62: Please check if all the yaml labels are the same for all clients.

config/dynconfig.go
```diff
@@ -0,0 +1,329 @@
...
+	// static config
+	ConfigInterval        *int `yaml:"config_interval"`
+	ConnectionQueueSize   *int `yaml:"connection_queue_size"`
```
khaf: Should be labeled as `max_connections_per_node`, same as Java. The labels should be unified.

config/dynconfig.go
mirzakaracic: The map to policy mappings will be only read if the file detects a change. Also we are not using map data structure in dynconfig.

```
	loadedConfig := dc.configProvider.LoadConfig(dc.dsn)
	if loadedConfig != nil {
	..
```

config/dynconfig.go
mirzakaracic: They are and where are discrepancies we convert map to that particular language expected property.

config.go
mirzakaracic: Yes so this was done to handle a case where users could enable dyn config however the configIntval value is either not set or the config object returned is nil. In that case we have to have a default value. The default value will be coming, on specified order, 1) from user passed in client policy 2) if no user policy provided use default client policy which defaults configInterval to 5 seconds.

bench_apply_configuration_test.go
mirzakaracic: I have left a comment but I think we are leaning toward static and dynamic sections still and have a version field on top of file.

mirzakaracic: There will be a version field in the file. Static and dynamic configuration will remain in place for now.

batch_read_policy.go
```diff
@@ -54,9 +56,21 @@ func NewBatchReadPolicy() *BatchReadPolicy {
...
-func (brp *BatchReadPolicy) toWritePolicy(bp *BatchPolicy) *WritePolicy {
+func NewDynamicBatchReadPolicy(dynConfig *DynConfig) *BatchReadPolicy {
+	if dynConfig == nil {
+		return NewBatchReadPolicy()
+	}
+
+	return dynConfig.client.dynDefaultBatchReadPolicy.Load()
+}
+
+func (brp *BatchReadPolicy) ToWritePolicy(bp *BatchPolicy, dynConfig *DynConfig) *WritePolicy {
```
khaf: No need to export this method.

batch_read_policy.go
```diff
@@ -54,9 +56,21 @@ func NewBatchReadPolicy() *BatchReadPolicy {
 	}
 }
 
-func (brp *BatchReadPolicy) toWritePolicy(bp *BatchPolicy) *WritePolicy {
+func NewDynamicBatchReadPolicy(dynConfig *DynConfig) *BatchReadPolicy {
```
khaf: No need to export.

batch_read_policy.go
```diff
@@ -66,5 +80,86 @@ func (brp *BatchReadPolicy) toWritePolicy(bp *BatchPolicy) *WritePolicy {
...
+		}
+		if dynConfig.config.Dynamic.BatchRead.ReadModeSc != nil {
+			brp.ReadModeSC = mapReadModeSCToReadModeSC(*dynConfig.config.Dynamic.BatchRead.ReadModeSc)
+		}
+	}
+
+	return brp
+}
+
+func (brp *BasePolicy) mapConfigBatchReadToBasePolicy(dynConfig *DynConfig) *BasePolicy {
```
khaf: This needs a better naming convention, or a logic rethinking.

batch_udf_policy.go
```diff
@@ -66,7 +66,15 @@ func NewBatchUDFPolicy() *BatchUDFPolicy {
 	}
 }
 
-func (bup *BatchUDFPolicy) toWritePolicy(bp *BatchPolicy) *WritePolicy {
+func NewDynamicBatchUdfPolicy(dynConfig *DynConfig) *BatchUDFPolicy {
```
khaf: Should not be exported.

batch_write_policy.go
```diff
@@ -93,9 +95,21 @@ func NewBatchWritePolicy() *BatchWritePolicy {
 	}
 }
 
-func (bwp *BatchWritePolicy) toWritePolicy(bp *BatchPolicy) *WritePolicy {
+func NewDynamicBatchWritePolicy(dynConfig *DynConfig) *BatchWritePolicy {
```
khaf: Should not be exported.

cluster.go
```diff
@@ -57,7 +57,7 @@ type Cluster struct {
 	// Hints for best node for a partition
 	partitionWriteMap iatomic.TypedVal[partitionMap] //partitionMap
 
-	clientPolicy        ClientPolicy
+	clientPolicy        iatomic.SyncVal[*ClientPolicy]
```
khaf: Should be an atomic pointer of `atomic.Pointer[ClientPolicy]`.

cluster.go
```diff
@@ -471,14 +473,15 @@ func (clstr *Cluster) waitTillStabilized() Error {
 		doneCh <- err
 	}()
 
+	clusterClientPolicy := *clstr.clientPolicy.Get()
```
khaf: No need to make a copy of the struct here.

config/registry/registry.go
```diff
@@ -0,0 +1,57 @@
...
+const (
+	DSN_REGEX_PATTERN = `^\s*(?P<scheme>[A-Za-z][A-Za-z0-9+.-]*://)?(?P<path>.*)$`
+	DEFAULT_SCHEME    = "file://"
+	DSN_SCHEME        = "scheme"
+	DSN_PATH          = "path"
+)
+
+var (
+	ConfigProvidersMu sync.RWMutex
+	ConfigProviders   = make(map[string]dynconfig.ConfigProvider)
```
khaf: Convert to `sync.Map`

logger/logger.go
```diff
@@ -57,6 +57,10 @@ func newLogger() *logger {
 	}
 }
 
+func (lgr *logger) IsLogLevelEnabled(level LogPriority) bool {
```
khaf: Check why this is used.

query_policy.go
```diff
@@ -56,3 +60,80 @@ func NewQueryPolicy() *QueryPolicy {
 		MultiPolicy: *NewMultiPolicy(),
 	}
 }
+
+func NewDynamicQueryPolicy(dynConfig *DynConfig) *QueryPolicy {
```
khaf: Do not export.

scan_policy.go
```diff
@@ -42,3 +46,82 @@ func NewScanPolicy() *ScanPolicy {
 		MultiPolicy: mp,
 	}
 }
+
+func NewDynamicScanPolicy(dynConfig *DynConfig) *ScanPolicy {
```
khaf: Do not export.

txn_roll_policy.go
```diff
@@ -35,3 +37,79 @@ func NewTxnRollPolicy() *TxnRollPolicy {
 		BatchPolicy: mp,
 	}
 }
+
+func NewDynamicTxnRollPolicy(dynConfig *DynConfig) *TxnRollPolicy {
```
khaf: Do not export.

txn_verify_policy.go
```diff
@@ -36,3 +38,79 @@ func NewTxnVerifyPolicy() *TxnVerifyPolicy {
 		BatchPolicy: mp,
 	}
 }
+
+func NewDynamicTxnVerifyPolicy(dynConfig *DynConfig) *TxnVerifyPolicy {
```
khaf: Do not export.

write_policy.go
```diff
@@ -103,3 +104,75 @@ func NewWritePolicy(generation, expiration uint32) *WritePolicy {
 
 	return res
 }
+
+func NewDynamicWritePolicy(dynConfig *DynConfig) *WritePolicy {
```
khaf: Do not export.

config.go
```diff
@@ -0,0 +1,535 @@
...
+	}
+
+	return config
+}
+
+// ----------------------------------------------------------------
+// Testing functions
+// ----------------------------------------------------------------
+
+func NewDynConfigForTest(config *dynconfig.Config) *DynConfig {
```
khaf: This should be moved to the `helper_test.go` file.

logger/logger.go
khaf: Not addressed yet.

khaf (changes requested):
A few trivial issues remain which need to be addressed.

mirzakaracic:
I am little confused by `outdated` comments on  changes. When I look at the code in the remote all of the comments have already been addressed except the last one? How are you reviewing the PR's?

khaf (approved):
Looks good.

PR #484, opened 2025-05-02
CLIENT-3438 Adding err to return

mirzakaracic:
Customer reported issue. Porting back to v7 since customer asked for it.


khaf (approved):

khaf:
Used cherry picking instead.

PR #488, opened 2025-05-07
CLIENT-3420 Update txn_verify_policy.go

ajkimdev:
Add SocketTimeout with 3 sec
Replace Millisecond to Second for align with Java client


mirzakaracic:
I have validated that the changes in the issue do address the issue. Also  crossed referenced with other clients that we do indeed use  the same defaults. AFAIK this should be good but will let @khaf decide since I might be overlooking something.

mirzakaracic (commented):
👍

mirzakaracic (approved):

PR #491, opened 2025-05-23
CLIENT-3406 Adding ability to pick select partition in PartitionFilter

mirzakaracic:


partition_filter.go
```diff
@@ -67,6 +69,38 @@ func NewPartitionFilterByKey(key *Key) *PartitionFilter {
...
+		if id < 0 || id >= _PARTITIONS {
+			return nil, newError(types.PARAMETER_ERROR, "Partition id out of range: %d", strconv.Itoa(id))
+		}
+
+		minPartitionId = min(minPartitionId, id)
+		maxPartitionId = max(maxPartitionId, id)
+	}
+	partitionCount := maxPartitionId - minPartitionId
+	partitionFilter := newPartitionFilter(minPartitionId, partitionCount)
+	slices.Sort(partitionIds)
```
khaf: Before sorting, this slice should be copied. Slices are pointers and the user owns this slice.

scan_test.go
```diff
@@ -315,6 +321,44 @@ var _ = gg.Describe("Scan operations", func() {
 		gm.Expect(len(keys)).To(gm.BeNumerically(">", 0))
 	})
 
+	gg.It("must Scan and get all partition records back for a specified partitions", func() {
```
khaf, line 323: This test needs to set the `policy.MaxRecords` to a small value to test for cursor resumption as well.

khaf (changes requested):
Looks good now. Just a trivial change to the user input and testing the cursor resumption and we are set.

scan_test.go
```diff
@@ -315,6 +321,47 @@ var _ = gg.Describe("Scan operations", func() {
...
+			partitions := getMasterPartitionIds(i)
+			scanPolicy.MaxRecords = 1
+			counter := 0
+
+			pf, _ := as.NewPartitionFilterSelectPartitions(partitions)
+			recordset, err := client.ScanPartitions(scanPolicy, pf, ns, set)
+
+			gm.Expect(err).ToNot(gm.HaveOccurred())
+
+			for res := range recordset.Results() {
```
khaf: `for rec, err := recordset.Records() {` will use an iterator, allows for more intuitive iteration and you will not forget to check for the error right after.

khaf (approved):
Looks good now.

PR #493, opened 2025-06-24
CLIENT-2226 - Changes for timeoutDelay

mirzakaracic:


cluster.go
```diff
@@ -79,6 +81,12 @@ type Cluster struct {
 
 	// Password in hashed format in bytes.
 	password iatomic.SyncVal[[]byte]
+
+	// Maximum socket idle to validate connections in command.
+	maxSocketIdleTran iatomic.SyncVal[*time.Duration]
```
khaf: As far as I can tell, there is no reason to have a pointer here.

cluster.go
```diff
@@ -79,6 +81,12 @@ type Cluster struct {
...
+
+	// Maximum socket idle to trim peak connections to min connections.
+	maxSocketIdleTrim iatomic.SyncVal[*time.Duration]
```
khaf: Same as above,

khaf (changes requested):
I don't see the recovery logic. I assume the PR was not ready for the review yet. We probably need a meeting to continue the progress.

mirzakaracic:
> I don't see the recovery logic. I assume the PR was not ready for the review yet. We probably need a meeting to continue the progress.

Mistake on my part. Was thinking we would be recovering, not closing connection, at executeAt but also need to handle the case when we attempt to `getConnection(....`.

batch_command.go
```diff
@@ -148,3 +150,7 @@ func (cmd *batchCommand) getNamespaces() iter.Seq2[string, uint64] {
 func (cmd *batchCommand) getNamespace() *string {
 	return &cmd.namespace
 }
+
+func (cmd *batchCommand) salvageConn(timeoutDelay time.Duration, conn *Connection, node *Node) {
```
khaf: If you define this on the `batchCommand`, it will automatically resolve in all the batch commands and you won't need to define it on every `inheriting` command.

cluster.go
```diff
@@ -109,11 +109,6 @@ func NewCluster(policy *ClientPolicy, hosts []*Host) (*Cluster, Error) {
 
 	clientPolicy := *policy
 
-	// Set a default Idle Timeout for the connection
```
khaf: Why?

command.go
```diff
@@ -3834,10 +3838,15 @@ func (cmd *baseCommand) executeAt(ifc command, policy *BasePolicy, deadline time
 			if deviceOverloadError(err) {
 				cmd.node.incrErrorCount()
 			}
+			// try to salvage the connection
+			if cmd.conn.salvageConnection {
```
khaf: Shouldn't `policy.TimeoutDelay` also be checked here? Maybe even turn it into a `func (ctn *connection) shouldSalvage(delay time.Duration)` so that it is never forgotten?

command.go
```diff
@@ -3871,9 +3880,15 @@ func (cmd *baseCommand) executeAt(ifc command, policy *BasePolicy, deadline time
 					}
 				}
 
-				// IO errors are considered temporary anomalies. Retry.
-				// Close socket to flush out possible garbage. Do not put back in pool.
-				cmd.conn.Close()
+				if cmd.conn.salvageConnection && policy.TimeoutDelay > 0 {
```
khaf, line 3883: Same as above if adding the method is elected.

connection.go
```diff
@@ -366,6 +379,8 @@ func (ctn *Connection) SetTimeout(deadline time.Time, socketTimeout time.Duratio
 
 // Close closes the connection
 func (ctn *Connection) Close() {
+	// Resetting lastUsed to zero time
```
khaf: Should be removed?

multi_command.go
```diff
@@ -158,11 +160,17 @@ func (cmd *baseMultiCommand) parseResult(ifc command, conn *Connection) Error {
 				return newError(types.PARSE_ERROR, fmt.Sprintf("Error setting up zlib inflater for size `%d`: %s", compressedSize-8, err.Error())).setNode(cmd.node)
 			}
 
+			// getting compressed received size
+			cmd.receiveSize += int64(receiveSize)
```
khaf: Is this a bug fix?

multi_command.go
```diff
@@ -158,11 +160,17 @@ func (cmd *baseMultiCommand) parseResult(ifc command, conn *Connection) Error {
...
+
 			// read the first 8 bytes
 			cmd.bc.reset(8)
 			if cmd.dataBuffer, err = cmd.bc.read(8); err != nil {
 				return err
 			}
+		} else {
+			// getting un-compressed received size
+			cmd.receiveSize += int64(receiveSize)
```
khaf: Same as above.

multi_command.go
```diff
@@ -509,3 +517,28 @@ func (cmd *baseMultiCommand) getNamespaces() iter.Seq2[string, uint64] {
 func (cmd *baseMultiCommand) getNamespace() *string {
 	return &cmd.namespace
 }
+
+func (cmd *baseMultiCommand) salvageConn(timeoutDelay time.Duration, conn *Connection, node *Node) {
+	// logger.Logger.Debug("TimeoutDelay enabled. Salvaging connection for node %s", node.GetName())
+	conn.deadline = time.Now().Add(timeoutDelay)
+	reader := bufio.NewReader(conn.conn)
+	discardedCount := int(conn.totalReceived - cmd.receiveSize)
```
khaf: This should be the other way around: `cmd.receiveSize - conn.totalReceived`

multi_command.go
```diff
@@ -509,3 +517,28 @@ func (cmd *baseMultiCommand) getNamespaces() iter.Seq2[string, uint64] {
...
+			if discarded < discardedCount {
+				conn.Close()
+				return
+			}
+		}
+		discardedCount -= discarded
+	}
+
+	node.PutConnection(conn)
+	conn.refresh()
```
khaf: Should should come before the line above.

node_test.go
```diff
@@ -184,175 +183,5 @@ var _ = gg.Describe("Aerospike Node Tests", func() {
 			})
 
 		})
-
-		gg.Context("When Idle Timeout Is Used", func() {
```
khaf: Why?

record_parser.go
```diff
@@ -63,6 +63,7 @@ func newRecordParser(cmd *baseCommand) (*recordParser, Error) {
 
 	// Read remaining message bytes.
 	receiveSize := int((sz & 0xFFFFFFFFFFFF))
+	//rp.cmd.receiveSize = int64(receiveSize)
```
khaf: Should be removed?

single_command.go
```diff
@@ -71,3 +72,28 @@ func (cmd *singleCommand) getNamespaces() iter.Seq2[string, uint64] {
 func (cmd *singleCommand) getNamespace() *string {
 	return &cmd.key.namespace
 }
+
+func (cmd *singleCommand) salvageConn(timeoutDelay time.Duration, conn *Connection, node *Node) {
+	// logger.Logger.Debug("TimeoutDelay enabled. Salvaging connection for node %s", node.GetName())
+	conn.deadline = time.Now().Add(timeoutDelay)
+	reader := bufio.NewReader(conn.conn)
+	discardedCount := int(conn.totalReceived - cmd.receiveSize)
```
khaf: Same as above.

khaf (changes requested):

multi_command.go
mirzakaracic: We are parsing one record at a time. Don't we need to read/add up receive size for all the records received, in order to get correct value that needs to be discarded?

multi_command.go
mirzakaracic: Left comment above.

command.go
mirzakaracic: > so that it is never forgotten?

I don't know about this. TimeoutDelay is part of dynamic config. If someone makes a change we need to take that into account. We could evaluate once per executeAt call and store results evaluate/check on write and read if connection should be salvaged. Otherwise we need to call `func (ctn *connection) shouldSalvage(delay time.Duration)` when policy is updated via dynamic config.

I would rather leave it the way it is, and add check for `policy.Timeoutdelaty > 0`,  and evaluate/re-evaluate on each call for simplicity reasons.

batch_command.go
mirzakaracic: Technically speaking we could move this function to baseCommand since the logic for most commands will be similar except for query and scan.

command.go
mirzakaracic: Left a comment [here](https://github.com/aerospike/aerospike-client-go/pull/493#discussion_r2250155746)

command.go
```diff
@@ -4162,3 +4184,28 @@ func (cmd *baseCommand) applyDetailedMetricsDataSizeAndLatencyOnWrite(ifc comman
 		}
 	}
 }
+
+func (cmd *baseCommand) salvageConn(timeoutDelay time.Duration, conn *Connection, node *Node) {
```
khaf: This implementation  does not work for streaming commands like Batch/Scan/Query. At least the batch case should be implemented.

connection.go
```diff
@@ -95,6 +97,10 @@ type Connection struct {
 	limitReader *io.LimitedReader
 
 	closer sync.Once
+
+	// Used to track the last time the connection was used. This is used to determine
```
khaf, line 101: The comment needs to be updated.

command.go
```diff
@@ -3834,10 +3839,15 @@ func (cmd *baseCommand) executeAt(ifc command, policy *BasePolicy, deadline time
 			if deviceOverloadError(err) {
 				cmd.node.incrErrorCount()
 			}
+			// try to salvage the connection
+			if cmd.conn.salvageConnection && policy.TimeoutDelay > 0 {
```
khaf, line 3842: No reason to check for salvage connection since it is market by `policy.TimeoutDelay anyway`.

khaf (changes requested):
The multi-message commands should be covered. Specifically batch in this case.

command.go
mirzakaracic: This is going to be done at a later point.

command.go
mirzakaracic: The idea here is that we want to handle only socket timeouts specifically.

khaf (approved):
Looks good.

PR #496, opened 2025-07-07
CLIENT-3452 - Circuit breaker progressive back off

mirzakaracic:


node.go
```diff
@@ -27,6 +27,11 @@ import (
 	"github.com/aerospike/aerospike-client-go/v8/types"
 )
 
+const (
+	MAX_ERROR_RATE_MIN_VALUE = 100
```
khaf, line 31: These need to be converted to CamelCase.

khaf: Also, users will want to set these independently and also change via dynamic config, so they should be in `ClientPolicy`.

khaf (changes requested):
Didn't zoom in on the logic of the back off, I assume it is correct by design.

The Max and Min error rates are something that users will want to be able to set independently for each cluster and change by dynamic config in the future, so I think they should be in the `ClientPolicy`.

node.go
mirzakaracic: The logic is following C implementation found [here](https://github.com/aerospike/aerospike-client-c/commit/83173e19b3258dadaac3688b791757470c893959).

> These need to be converted to CamelCase.

I am fine with converting to camelcase but we are not using CamelCase with other constants in the same file.

> Also, users will want to set these independently and also change via dynamic config, so they should be in ClientPolicy.

The users will be able to set MaxErrorRate and ErrorRateWindow values using the dynamic config. This is already available as part of the dynamic config PR. The ratio of ErrorRateWindow to MaxErrorRate 1:100 is set as the max ratio that is allowed. Users are able to configure ratios that lower but not greater. For instance user will not be allowed to set MaxErrorRate to 200 and leave ErrorRateWindow to 1. This has been outlined in the following [doc](https://aerospike.atlassian.net/wiki/x/AQDm-g)

khaf (approved):
Looks good.

PR #498, opened 2025-07-21
CLIENT-3571 - Changes to handle errors for PKI users

mirzakaracic:


client.go
```diff
@@ -1840,8 +1840,8 @@ func (clnt *Client) DropUser(policy *AdminPolicy, user string) Error {
 func (clnt *Client) ChangePassword(policy *AdminPolicy, user string, password string) Error {
 	policy = clnt.getUsableAdminPolicy(policy)
 
-	if clnt.cluster.user == "" {
```
khaf: The check is for making sure there is a valid logged-in user represented in the client.

khaf (changes requested):

khaf (approved):
Looks good.

PR #499, opened 2025-07-24
CLIENT-3590 - Secondary index changes

mirzakaracic:


mirzakaracic:
The tests are failing since we need AES version 8.1.0.0. When testing locally against 8.1.0.0-rc2 tests are passing. Will add `server-version` to AES github action to get the tests to run against correct version of AES DB.

https://github.com/aerospike/aerospike-client-go/pull/499/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R27

.github/workflows/build.yml
```diff
@@ -23,6 +23,8 @@ jobs:
         run: go version
       - name: Set up Aerospike Database
         uses: reugn/github-action-aerospike@v1
+        with:
+          server-version: 8.1.0.0-rc4_1
```
khaf: Not necessarily for this PR, but can we have a matrix here?

client.go
```diff
@@ -1539,6 +1539,39 @@ func (clnt *Client) CreateComplexIndex(
...
+// This method is only supported by Aerospike 3+ servers.
+// If the policy is nil, the default relevant policy will be used.
+func (clnt *Client) CreateIndexWithExpression(
+	policy *WritePolicy,
+	namespace string,
+	setName string,
+	indexName string,
+	indexType IndexType,
+	indexCollectionType IndexCollectionType,
+	expression *Expression,
```
khaf, line 1559: No `ctx` for expression indexes?

khaf (dismissed):
Looks Good. No `ctx` for expression filters?

client.go
mirzakaracic: No this is what was asked for explicitly. Java and other clients have something similar

https://github.com/aerospike/aerospike-client-java/blob/6d8fbbc118bde5c7e292ba91cfe029ded244e651/client/src/com/aerospike/client/AerospikeClient.java#L4824

mirzakaracic:
> No ctx for expression filters?

Looking at what has been done for Java and other clients it seems that is not what is asked for.

[java](https://github.com/aerospike/aerospike-client-java/blob/6d8fbbc118bde5c7e292ba91cfe029ded244e651/client/src/com/aerospike/client/query/Filter.java#L33)
[c#](https://github.com/aerospike/aerospike-client-csharp/blob/b62fd5f3c2267268eed49b36f68dfc512e36b1a6/AerospikeClient/Query/Filter.cs#L17)

command.go
```diff
@@ -2751,6 +2766,17 @@ func (cmd *baseCommand) setQuery(policy *QueryPolicy, wpolicy *WritePolicy, stat
 				return newCommonError(err)
 			}
 		}
+
+		if statement.Filter.indexName != "" {
```
khaf, line 2770: `if strings.TrimSpace(statement.Filter.indexName) != "" {`

khaf (approved):
Looks good.

PR #500, opened 2025-07-25
CLIENT-3546 Added changes to fall in line with logging messages by other clients

mirzakaracic:
The changes in this PR encompass following items
- Changes requested by product as part of CLIENT-3546
- Removed global lock on dynConfig
- Bug fixes encountered during testing

### DIFF
--- a/config.go
+++ b/config.go
@@ -1,4 +1,4 @@
-// Copyright 2014-2022 Aerospike, Inc.
+// Copyrigh 2014-2022 Aerospike, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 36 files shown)


config.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2014-2022 Aerospike, Inc.
+// Copyrigh 2014-2022 Aerospike, Inc.
```
khaf: ?

khaf (approved):
Looks good.

config.go
mirzakaracic: Good catch! I blame copy paste :)

khaf (dismissed):
Looks good.

khaf (approved):
Way too huge to meaningfully screen. A cursory look didn't immediately reveal any issues.

PR #502, opened 2025-08-11
CLIENT-3586 - user-agent changes

mirzakaracic:


node.go
```diff
@@ -449,8 +451,7 @@ func (nd *Node) newConnectionAllowed() Error {
 	return nil
 }
 
-// newConnection will make a new connection for the node.
-func (nd *Node) newConnection(overrideThreshold bool) (*Connection, Error) {
+func (nd *Node) newConnectionWithDetail(overrideThreshold bool, isTendConn bool) (*Connection, Error) {
```
khaf: This is leaky design. No need to do everything in the same function. Rename it back to what it was, make new one:
```go
func (no *Node) newTendConnection() (*Connection, Error) {
    conn, err := nd.newConnection(true, true);
    if err != nil {
          return nil, err
    }

    if err := nd.sendUserAgentId(conn); err != nil {
        // If setting user agent failed, we still return the connection
        // as it is already authenticated and usable.
        logger.Logger.Warn("Error setting user agent for node %s: %s", nd.String(), err.Error())
    }

    return conn, nil
}
```

node.go
```diff
@@ -505,9 +506,23 @@ func (nd *Node) newConnection(overrideThreshold bool) (*Connection, Error) {
 	nd.stats.ConnectionsSuccessful.IncrementAndGet()
 	conn.setIdleTimeout(clusterClientPolicy.IdleTimeout)
 
+	if isTendConn {
```
khaf: Remove these since they have moved into the new function.

node.go
```diff
@@ -715,7 +730,7 @@ func (nd *Node) usingTendConn(timeout time.Duration, f func(conn *Connection)) (
 			if nd.connectionCount.Get() == 0 {
 				// if there are no connections in the pool, create a new connection synchronously.
 				// this will make sure the initial tend will get a connection without multiple retries.
-				*conn, err = nd.newConnection(true)
+				*conn, err = nd.newConnectionWithDetail(true, true)
```
khaf: Here call the new `makeTendConn` function.

khaf (changes requested):

cluster.go
```diff
@@ -1067,3 +1076,26 @@ func (clstr *Cluster) getNodeLabels(metricPolicy *MetricsPolicy) *Labels {
...
 }
+
+//-------------------------------------------------------
+// Utility Functions
+//-------------------------------------------------------
+
+// getLibraryVersion returns the version of the aerospike client library.
+func getLibraryVersion(modulePath string) string {
+	// golang idiomatic module version is "(devel)" if not set.
+	defaultVersion := "(devel)"
```
khaf, line 1087: Should include a `Go Client` identifier somewhere in there. Also for other clients.

cluster.go
```diff
@@ -1067,3 +1076,26 @@ func (clstr *Cluster) getNodeLabels(metricPolicy *MetricsPolicy) *Labels {
...
+
+	info, ok := debug.ReadBuildInfo()
+	if !ok {
+		return defaultVersion
+	}
+
+	for _, dep := range info.Deps {
+		if dep.Path == modulePath {
+			return dep.Version
```
khaf, line 1096: Do we have anything identifying the Go client here? If not, should be added.

khaf (changes requested):

cluster.go
mirzakaracic: So the way this works we call getLIbraryVersion(...) when we instantiate cluster instance [here](https://github.com/aerospike/aerospike-client-go/blob/979aba25fa96e292c0143beddb7713b39975e123/cluster.go#L141). The value is module name we are looking to get version from. Using debug.ReadBuildInfo() to get module version.

The value that is passed in is a constant https://github.com/aerospike/aerospike-client-go/blob/979aba25fa96e292c0143beddb7713b39975e123/cluster.go#L34 

> Should include a Go Client identifier somewhere in there. Also for other clients.

The full userAgenId is constructed [here](https://github.com/aerospike/aerospike-client-go/blob/979aba25fa96e292c0143beddb7713b39975e123/node.go#L1003)

Sample output from cluster that has three nodes and go and java clients attached to the cluster

```bash
Admin> show user-agents
~User Agent Information (2025-08-19 18:33:31 UTC)~
            Node|     Client| App ID|Count
                |    Version|       |
10.88.0.214:3100|asadm-4.0.0|unknown|    1
10.88.0.214:3100|go-v8.0.1  |not-set|    1
10.88.0.214:3100|java-18.0.0|null   |    1
10.88.0.215:3101|asadm-4.0.0|unknown|    1
10.88.0.215:3101|go-v8.0.1  |not-set|    1
10.88.0.215:3101|java-18.0.0|null   |    1
10.88.0.216:3102|asadm-4.0.0|unknown|    1
10.88.0.216:3102|go-v8.0.1  |not-set|    1
10.88.0.216:3102|java-18.0.0|null   |    1
```

cluster.go
mirzakaracic: Answer is in https://github.com/aerospike/aerospike-client-go/pull/502#discussion_r2285843567

cluster.go
khaf: `debug.ReadBuildInfo()` only gives the information for the main module of an executable.

khaf (changes requested):

khaf (approved):
After talking to Mirza, it seems the `debug.ReadBuildInfo()` works as expected per his tests.

PR #504, opened 2025-08-20
CLIENT-3120 - Replace an existing node in the cluster when a new peer has the same node name, but a different IP address

mirzakaracic:
## Changes
**Bug fix:**
Used https://aerospike.atlassian.net/browse/CLIENT-3117 as and [java commit](https://github.com/aerospike/aerospike-client-java/commit/38585450f0f37937dc1cef6ec4b06c83ecb8eff3) as reference for the changes. 

This ticket also takes care of https://aerospike.atlassian.net/browse/CLIENT-3318 since we have added logic to prevent same node being added to remove list if the node is already on the list.
https://github.com/aerospike/aerospike-client-go/blob/c6ffa3e01cc53d3d79c2f07545d9be0577a88cd3/cluster.go#L316

**Validation:**
To validate changes I have used instructions posted [here](https://aerospike.atlassian.net/browse/OPS-7572). 

I was able to validate that on ip change the client was able to update node ipaddress to new ip.

```bash
Removed nodes: [BB93A00800A0142 34.44.198.103:3000]
2025/08/25 21:58:45 benchmark.go:813: write(tps=1 timeouts=0 errors=0) read(tps=0 timeouts=0 errors=0) total(tps=1 timeouts=0 errors=0, count=352)
```

**Added feature:**
Added ability to use service-alternate when using benchmark utility.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -286,7 +286,7 @@ func (clstr *Cluster) tend() Error {
 
 	// find the first host that connects
 	seq.ParDo(peers.peers(), func(_peer *peer) {
-		if clstr.peerExists(peers, _peer.nodeName) {
+		if clstr.peerExists(peers, _peer) {
 			// Node already exists. Do not even try to connect to hosts.
 			return
 		}
@@ -304,7 +304,7 @@ func (clstr *Cluster) tend() Error {
 				logger.Logger.Warn("Peer node `%s` is different than actual node `%s` for host `%s`", _peer.nodeName, nv.name, host)
 			}
 
-			if clstr.peerExists(peers, nv.name) {
+			if clstr.peerExists(peers, _peer) {
 				// Node already exists. Do not even try to connect to hosts.
 				return seq.Break
 			}
@@ -315,6 +315,10 @@ func (clstr *Cluster) tend() Error {
 			partMap.InitDoVal(clstr.getPartitions().clone, func(partMap partitionMap) {
 				node.refreshPartitions(peers, partMap, true)
 			})
+
+			if _peer.replaceNode != nil {
+				peers._nodesToRemove = append(peers._nodesToRemove, _peer.replaceNode)
+			}
 			return seq.Break
 		})
 	})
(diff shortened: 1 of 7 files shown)


peers.go
```diff
@@ -21,21 +21,23 @@ import (
...
 func newPeers(peerCapacity int, addCapacity int) *peers {
 	return &peers{
-		_peers:     make(map[string]*peer, peerCapacity),
-		_hosts:     make(map[Host]struct{}, addCapacity),
-		_nodes:     make(map[string]*Node, addCapacity),
-		genChanged: *atomic.NewBool(true),
+		_peers:         make(map[string]*peer, peerCapacity),
+		_hosts:         make(map[Host]struct{}, addCapacity),
+		_nodes:         make(map[string]*Node, addCapacity),
+		_nodesToRemove: make([]*Node, 0),
```
khaf, line 40: Idiomatically can be `make([]*Node)`.

cluster.go
```diff
@@ -304,17 +304,22 @@ func (clstr *Cluster) tend() Error {
...
 			})
+
+			if _peer.replaceNode != nil {
+				// Preventing duplicate entries.
+				for _, entry := range peers._nodesToRemove {
+					if entry.Equals(_peer.replaceNode) {
+						return seq.Break
+					}
+				}
+				peers._nodesToRemove = append(peers._nodesToRemove, _peer.replaceNode)
```
khaf: `seq.ParDo` does stuff concurrently for all the list elements, so there is a potential race condition here. We probably need a `addNodesToRemove()` method on `peers` struct, and `peers._nodesToRemove` should be moved into a `SyncVal` or something. The above logic can also be moved into that method, simplifying the code here.

khaf (changes requested):
I believe we have a race condition here that needs to be addressed.

peers.go
mirzakaracic: Not sure I follow. We are using a slice here and capacity is required. Unless you are saying we should use a map.

khaf: That's the length and not capacity. `make([]*Node)` would make a map with length 0 and predetermined capacity as well, same as here.

khaf (dismissed):
Looks good.

PR #510, opened 2025-09-08
CLIENT-3435 Update Go Client to version 1.24 to support FIPS

mirzakaracic:
Validated changes locally as per https://go.dev/doc/security/fips140.


khaf (changes requested):
Changing the required Go version is somewhat of a breaking change.

If you are going to do it, then I think you should go to the highest, which is v1.25.1 at the moment.

khaf (approved):
Looks good.

PR #511, opened 2025-09-09
CLIENT-3021 Statement.BinNames must be empty for QueryExecute

mirzakaracic:
Used java as base. More specifically the [execute](https://github.com/aerospike/aerospike-client-java/blob/aaf1f793886b5ff8852e3854f860aaed23bc0a02/client/src/com/aerospike/client/AerospikeClient.java#L4237) method on java side does not check and throw an exception if statement.binNames is not empty.

### DIFF
--- a/error.go
+++ b/error.go
@@ -378,7 +378,6 @@ var (
 	ErrNetTimeout                      = newConstError(types.TIMEOUT, "network timeout")
 	ErrUDFBadResponse                  = newConstError(types.UDF_BAD_RESPONSE, "invalid UDF return value")
 	ErrNoOperationsSpecified           = newConstError(types.INVALID_COMMAND, "no operations were passed to QueryExecute")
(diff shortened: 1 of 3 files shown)


khaf (dismissed):
Looks Good.

error.go
```diff
@@ -378,7 +378,6 @@ var (
 	ErrNetTimeout                      = newConstError(types.TIMEOUT, "network timeout")
 	ErrUDFBadResponse                  = newConstError(types.UDF_BAD_RESPONSE, "invalid UDF return value")
 	ErrNoOperationsSpecified           = newConstError(types.INVALID_COMMAND, "no operations were passed to QueryExecute")
-	ErrNoBinNamesAllowedInQueryExecute = newConstError(types.INVALID_COMMAND, "`Statement.BinNames` must be empty for QueryExecute")
```
khaf, line 381: Breaking change. Should not be removed without major version update.

khaf (changes requested):

khaf (approved):
Looks good.

PR #513, opened 2025-09-15
CLIENT-3023 v8 Allow repeat background query with the same statement

mirzakaracic:
Changes for v8 branch for CLIENT-3023


khaf (approved):
Looks good.

PR #517, opened 2025-09-26
Revert module go version to 1.23

reugn:
Although the `crypto/fips140` module was introduced with go 1.24, the client is not required to target go 1.24 or higher. The FIPS Compliance test can be conditioned using a build constraint.

### DIFF
--- a/fips_compliance_test.go
+++ b/fips_compliance_test.go
@@ -1,3 +1,5 @@
+//go:build go1.24
+
 // Copyright 2014-2022 Aerospike, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


mirzakaracic:
> the client is not required to target go 1.24 or higher

The client is not required however it does make the process on validation a lot easier since go lang builtin toolchain can be used. 

Any specific reasons why this is a concern?

reugn:
The FIPS validation doesn't require any features specific to the newer Go version, so it would unnecessarily restrict users who are still on go1.23. CI will verify fips140 applicability by testing using a newer Go version.

khaf (approved):

PR #518, opened 2025-09-29
Eliminate dynamic configuration watcher goroutine leak

reugn:
* Eliminate dynamic configuration watcher goroutine leak.
* Remove unused code and format.
* Fix typos.


mirzakaracic:
> Eliminate dynamic configuration watcher goroutine leak.

Under what conditions did you run into the leak? The reason I am asking is because we using a similar approach for tend cycle.

reugn:
`dc.configWatchChannel` is never closed, which leaves the watcher goroutine stuck in the for loop even after the client is closed.

config/provider/yaml_config_provider.go
```diff
@@ -45,12 +45,6 @@ func NewYamlConfigProvider() dynconfig.ConfigProvider {
 	return &YamlConfigProvider{oldModTime: time.Time{}}
 }
 
-func NewYamlConfigProviderWithPath(configFilePath string) dynconfig.ConfigProvider {
```
khaf, line 48: This is a breaking change for a minor release. Should be returned.

khaf (changes requested):

config/provider/yaml_config_provider.go
reugn: Reverted, but it can be safely removed in the next major release.

khaf (approved):
Looks good.

PR #519, opened 2025-10-08
CLIENT-3796 - Background query for all records applies only partially

mirzakaracic:
This fixes the issue that was introduced by converting _nodes from map[string, *Node] to atomic.Map[string, *Nodes]. Essentially we are running into a stale data issue which is manifesting itself when using Clone() on nodes() when using atomic.Map[string, *Node].

### DIFF
--- a/peers.go
+++ b/peers.go
@@ -15,6 +15,7 @@
 package aerospike
 
 import (
+
 	"github.com/aerospike/aerospike-client-go/v8/internal/atomic"
 	sm "github.com/aerospike/aerospike-client-go/v8/internal/atomic/map"
 )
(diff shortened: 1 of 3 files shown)


peers.go
```diff
@@ -68,7 +75,9 @@ func (ps *peers) peers() []*peer {
 
 // nodes returns a copy of nodes for safe iteration
 func (ps *peers) nodes() map[string]*Node {
-	return ps._nodes.Clone()
```
khaf, line 71: You can just implement `.AsRef() ` and return the underlying map. But probably you should check where this function is used and make it use a map function instead of direct iteration.

khaf (changes requested):

khaf (approved):
Looks good.

PR #522, opened 2025-10-15
fix unsafe behaviour in connection.go:errToAerospikeErr() 

equals215:
In `func errToAerospikeErr(conn *Connection, err error) (aerr Error)` the provided conn may be salvaged by modifying one of the fields if the error originates from a `net` Timout or DeadlineExceeded.  
However in `func newConnection(address string, timeout time.Duration) (*Connection, Error)` if the `net.DialTimeout()` call timeouts, the non-nil err branch calls `errToAerospikeErr(nil, err)` ; this then leads to a `SIGSEGV`.  

If you wanna reproduce just have a distant Aerospike cluster behind some kind of poor networking (Tailscale should do the trick) and connect to it using the go client.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -121,7 +121,7 @@ func errToAerospikeErr(conn *Connection, err error) (aerr Error) {
 			if conn != nil && conn.node != nil {
 				conn.node.stats.ConnectionsTimeoutErrors.IncrementAndGet()
 			}
-			if errors.Is(terr, os.ErrDeadlineExceeded) {
+			if conn != nil && errors.Is(terr, os.ErrDeadlineExceeded) {
 				conn.salvageConnection = true
 			}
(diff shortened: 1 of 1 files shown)


mirzakaracic:
@equals215 Thank you for reporting the issue. We are aware of it internally and will have a fix soon.

equals215:
> @equals215 Thank you for reporting the issue. We are aware of it internally and will have a fix soon.

Isn't that a fix already? Or this suggest there are deeper problems that I'm not aware of?

mirzakaracic:
> Isn't that a fix already? 

There is more to it. :)

equals215:
>There is more to it. :)

Can't wait for the fix, any change you tag this PR in a way I get a notification so I don't have to use my fork anymore?

Closing the PR

PR #523, opened 2025-10-15
CLIENT-3810 nil pointer panic occurs on errToAerospikeErr

mirzakaracic:


command.go
```diff
@@ -3871,6 +3871,7 @@ func (cmd *baseCommand) executeAt(ifc command, policy *BasePolicy, deadline time
 				// IO errors are considered temporary anomalies. Retry.
 				// Close socket to flush out possible garbage. Do not put back in pool.
 				cmd.conn.Close()
+				cmd.conn = nil
 			}
 
 			cmd.conn = nil
```
khaf, line 3876: Already set here to nil.

command.go
khaf: This line can hit a nil deref error due to line [#3918](https://github.com/aerospike/aerospike-client-go/pull/523/files#diff-4d7d1923693fc5ce892add2ea2907a744e77ea0b50c1939ccc5067cb48a466a3R3918)

single_command.go
```diff
@@ -74,7 +74,13 @@ func (cmd *singleCommand) getNamespace() *string {
 }
 
 func (cmd *singleCommand) salvageConn(timeoutDelay time.Duration, conn *Connection, node *Node) {
+	// If the connection is already closed, don't bother trying to salvage it.
+	if !cmd.conn.IsConnected() {
```
khaf: Shouldn't `cmd.conn != nil` also be added as a condition here?

khaf (changes requested):

khaf (approved):
Looks good.

command.go
khaf: You have a potential `cmd.conn = nil` issue here.

PR #527, opened 2025-10-20
fix nil point dereference error

back-D:


mirzakaracic:
Thank you for your contribution @back-D. The fix is merged in stage.
https://github.com/aerospike/aerospike-client-go/commit/6ab5438bffaeca821d5f76b14a21deb38a0c5cb8

PR #531, opened 2025-10-29
CLIENT-3847 update go ci workflow

gagan405:
* Add server version to build workflow - dynamically generate the build matrix
* go version is taken from `go.mod`
* verified in my fork that the builds are working as expected

I haven't added any server-version file yet.


.github/workflows/build.yml
```diff
@@ -8,14 +8,40 @@ env:
...
+    timeout-minutes: 30
+    needs: make-matrix
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-version:
-          - 1.25.1
-        server-version:
-          - 8.1.0.0
+        include: ${{ fromJson(needs.make-matrix.outputs.input-matrix) }}
```
gagan405, line 44: This actually brings down the build time go version to 1.23 as mentioned in go.mod. Updating the version in go.mod results in errors - which can be taken up as a separate task.

mirzakaracic (approved):
👍 Looks good

PR #532, opened 2025-10-30
CLIENT-3749 Path Expressions

mirzakaracic:
Server image for 8.1.1 is not available yet. `build.yaml` is not updated to test against that particular image yet because it it not available yet. The tests for pathExpressions are skipped at the moment since we are checking for server version

https://github.com/aerospike/aerospike-client-go/blob/77f52e501ce558c0617edc6cc4326c53fb03043c/cdt_operation_test.go#L44

https://github.com/aerospike/aerospike-client-go/blob/75f54f61ce9a1e227e9493b54bfe276a69e46c96/exp_cdt_test.go#L43


cdt_context.go
```diff
@@ -23,22 +23,24 @@ import (
...
 // CDTContext defines Nested CDT context. Identifies the location of nested list/map to apply the operation.
 // for the current level.
 // An array of CTX identifies location of the list/map on multiple
 // levels on nesting.
 type CDTContext struct {
-	Id    int
-	Value Value
+	Id         int
+	Value      Value
+	Expression *Expression
```
gagan405, line 43: nit: From the logic below, I infer that either `Value` or `Expression` can be present, but not both. And `Expression` takes priority over `Value`(?). Could be helpful to update the documentation/comments here.

cdt_list.go
```diff
@@ -294,10 +296,19 @@ func packCDTIfcVarParamsAsArray(packer BufferEx, opType int16, ctx []*CDTContext
...
+					return size + n, err
+				}
+				size += n
+			} else if c.Expression != nil {
+				if n, err = c.Expression.pack(packer); err != nil {
+					return size + n, err
+				}
+				size += n
+			} else {
+				return size, newError(types.PARAMETER_ERROR, "CDTContext must have either a Value or an Expression")
```
gagan405, line 315: should we also raise an error if both `Value` and `Expression` are present? If they are mutually exclusive, I think an error will be helpful here when both of them are present.

expression.go
```diff
@@ -1426,3 +1774,79 @@ func ExpUnknown() *Expression {
...
+		return ExpIntVal(int64(v))
+	case float32:
+		return ExpFloatVal(float64(v))
+	case float64:
+		return ExpFloatVal(v)
+	case string:
+		return ExpStringVal(v)
+	case []byte:
+		return ExpBlobVal(v)
+	case []interface{}:
```
khaf: This should be removed from here and moved to a generic function.

expression.go
```diff
@@ -1426,3 +1774,79 @@ func ExpUnknown() *Expression {
...
+			_, err := packObject(packer, e, false)
+			if err != nil {
+				panic(err)
+			}
+			return &Expression{bytes: packer.Bytes()}
+		}
+		return ExpListValueVal(v)
+	case map[string]interface{}:
+		return newFilterExpression(nil, JsonValue(v), nil, nil, nil, nil)
+	case map[interface{}]interface{}:
```
khaf: Should be removed from here and moved to a generic function.

expression.go
```diff
@@ -1426,3 +1774,79 @@ func ExpUnknown() *Expression {
...
+		return ExpMapVal(MapValue(v))
+	default:
```
khaf: Not sure this behavior should be the default. We will discuss this in a meeting.

expression.go
```diff
@@ -1426,3 +1774,79 @@ func ExpUnknown() *Expression {
 		arguments: nil,
 	}
 }
+
+func NewExpression(e interface{}) *Expression {
```
khaf: If this is intended to be used as a sugar, the name is too long.

khaf (changes requested):
Seems ok to me, but let's discuss a few minor issues before merge.

khaf (approved):

PR #534, opened 2025-11-04
Fix for panic if requestInfo call fails

mirzakaracic:


client.go
```diff
@@ -1861,11 +1861,17 @@ func (clnt *Client) CreateUser(policy *AdminPolicy, user string, password string
 		return err
 	}
 
-	node.usingTendConn(policy.Timeout, func(conn *Connection) {
+	errCall := node.usingTendConn(policy.Timeout, func(conn *Connection) Error {
```
khaf: You can just return the call now:
```go
return node.usingTendConn(...);

File: client.go [hunk 2]
```diff
@@ -1861,11 +1861,17 @@ func (clnt *Client) CreateUser(policy *AdminPolicy, user string, password string
...
 		command := NewAdminCommand(nil)
 		err = command.createUser(conn, policy, user, hash, roles)
+
+		return err
 	})
 
+	if errCall != nil {
```
khaf: And remove all the other returns.

client.go
```diff
@@ -1861,11 +1861,17 @@ func (clnt *Client) CreateUser(policy *AdminPolicy, user string, password string
...
+		return errCall
+	}
+
 	return err
```
khaf: Same as this.

client.go
```diff
@@ -1891,11 +1897,17 @@ func (clnt *Client) CreatePKIUser(policy *AdminPolicy, user string, roles []stri
 		return newCommonError(nil, fmt.Sprintf("Node version %s is less than required minimum version %s", node.version.String(), serverMinVersion))
 	}
 
-	node.usingTendConn(policy.Timeout, func(conn *Connection) {
+	errCall := node.usingTendConn(policy.Timeout, func(conn *Connection) Error {
 		command := NewAdminCommand(nil)
 		err = command.createUser(conn, policy, user, hash, roles)
+
+		return err
```
khaf: Remove.

client.go
```diff
@@ -1891,11 +1897,17 @@ func (clnt *Client) CreatePKIUser(policy *AdminPolicy, user string, roles []stri
...
 	})
 
+	if errCall != nil {
```
khaf: Also remove.

client.go
```diff
@@ -1939,7 +1958,7 @@ func (clnt *Client) ChangePassword(policy *AdminPolicy, user string, password st
 		return err
 	}
 
-	node.usingTendConn(policy.Timeout, func(conn *Connection) {
+	errCall := node.usingTendConn(policy.Timeout, func(conn *Connection) Error {
```
khaf: `err := ...`

client.go
khaf: `err := ...`

client.go
```diff
@@ -1985,11 +2017,17 @@ func (clnt *Client) RevokeRoles(policy *AdminPolicy, user string, roles []string
...
```
khaf: Same as above. Both returns should be removed.

PR #537, opened 2025-11-10
Core 1478

gburanov:


policy.go
```diff
@@ -148,6 +148,8 @@ type BasePolicy struct {
 	// Scan and query are also not affected by replica algorithms.
 	// Default to sending read commands to the node containing the key's master partition.
 	ReplicaPolicy ReplicaPolicy
+
+	ExitFastOnExhaustedConnectionPool bool
```
gburanov, line 152: I think it make sense to use this feature https://github.com/remerge/aerospike-client-go/commit/205ba61dd73f175cd05092d7fc601d485080212b but with new naming `ExitFastOnExhaustedConnectionPool`

So when we will use even newer aerospike version - we will just drop this commit and it will work

PR #538, opened 2025-11-13
CLIENT-3893 Add synchronization to `Connection.net.Conn` to prevent data races and panics

reugn:
A nil pointer panic was encountered during tests combining multiple scan operations and parallel node restarts.

The panic arises because multiple operations share a connection (e.g. tend) that can be closed prematurely if an individual operation fails.

The `net.Conn` property in the `Connection` struct lacks synchronization, leading to potential `Read` or `Write` calls on a nil connection.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -74,7 +74,9 @@ type Connection struct {
 	idleDeadline time.Time
 
 	// connection object
-	conn          net.Conn
+	conn net.Conn
+	// connection mutex
+	connMu        sync.RWMutex
 	totalReceived int64
 
 	// histogram to adjust the buff size to optimal value over time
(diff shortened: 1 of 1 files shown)


mirzakaracic (changes requested):
The PR needs to be created against stage branch not v8. Other than that it looks good. Thank you @reugn .

reugn:
@mirzakaracic, the base branch has been changed to stage.

mirzakaracic (approved):

khaf:
This change should be reversed. Please do not merge changes like this without my review.

mirzakaracic:
@reugn please send us a stack trace so we can understand where the issue is happening exactly.

PR is open to revert the change. Will reopen PR with same set of changes as this PR.

PR #539, opened 2025-11-13
Revert "CLIENT-3893 Add synchronization to `Connection.net.Conn` to prevent data races and panics"

mirzakaracic:
Reverts aerospike/aerospike-client-go#538


khaf (approved):
Looks good.

PR #545, opened 2025-11-20
CLIENT-3891 TaskId not being properly assigned during Background Query

mirzakaracic:


khaf:
@mirzakaracic It seems to me that this PR is unnecessary. The bug is in the `statement.prepareTaskId` method, where `stmt.TaskId` is not set. It looks like that method has changed meaning over time and needs a bit of dust off. We can discuss this in a short meeting.

mirzakaracic:
So the issue is that we default the taskId in the statement to 0. Now we are settings the taskId 

https://github.com/aerospike/aerospike-client-go/blob/e026ba54aeec272a4e6130eaac1ae8b9fb9ba489/client.go#L1377

however the taskId that is used in now propagated on `command := newServerCommand(`. And the issue that this fix aimes to address is when when make the following call

https://github.com/aerospike/aerospike-client-go/blob/e026ba54aeec272a4e6130eaac1ae8b9fb9ba489/server_command.go#L35

In this case if we don't update the taskId or pass down the taskId that has been prepared on initial QueryExecute call we are going write buffer with taskId 0. 

Later when someone tries to query for that job using info calls they will not be able to. The test case that was added exercises this exact scenario.

If the use case is not valid I take it we don't need to make any changes but if it's valid we should take this fix.

khaf:
Talked to @mirzakaracic about the current defective design in the `statement.taskId`. We'll add some deprecation warnings to later change the design for a proper API for it.

khaf (approved):
Looks good.

PR #547, opened 2025-11-26
CLIENT-3925 - Avoid sending previous errors when cluster is complete

reugn:
ABS resilience tests revealed an issue where the client sends previously aggregated errors even after a scan successfully completes.

This PR implements a fix to resolve the issue.


khaf (approved):
Looks good.

PR #562, opened 2025-12-08
CLIENT-3090: added document demonstrating the async way of using client api to invoke server ops

dmishra-aero:
Added examples to showcase Single Async Operation, Batch Async Operation that a enduser could do with Go client Api

### DIFF
--- a/docs/async_api_usage.md
+++ b/docs/async_api_usage.md
@@ -15,7 +15,7 @@ Unlike some other language clients, the Go client doesn't have explicit async/aw
 
 Before using async operations, ensure you have:
 
-1. **Go 1.18+** installed
+1. **Go 1.24.0** installed
 2. **Aerospike Go client v8** installed:
    ```bash
    go get github.com/aerospike/aerospike-client-go/v8
@@ -73,7 +73,7 @@ Single async operations let you execute individual Put/Get calls independently. 
 
 **Note**: A single Get/Put call does not require a goroutine; it can be invoked directly. The provided examples use goroutines only to show how these operations can participate in concurrent workflows.
 
-### Write Operations (Put)
+### (Put) Operations 
 
 The following example demonstrates how to write singe records:
 
@@ -142,7 +142,7 @@ func asyncPut(client *as.Client, policy *as.WritePolicy, key *as.Key, bins ...*a
 
 ```
 
-### Read Operations (Get)
+### (Get) Operations
 
 The following example demonstrates how to read single records:
 
@@ -214,7 +214,7 @@ func asyncGet(client *as.Client, policy *as.BasePolicy, key *as.Key, binNames ..
   
 ```
 
-### Operate Operations
+### (Operate) Operations
 
 You can also perform async Operate operations (combining read and write):
 
@@ -292,9 +292,9 @@ func asyncOperate(client *as.Client, policy *as.WritePolicy, key *as.Key, ops ..
 
 ## Batch Async Operations
 
-Batch operations in the Aerospike Go client already use internal concurrency (via goroutines) to process multiple nodes in parallel. However, you can run multiple batch operations concurrently for even better performance.
+However, you can run multiple batch operations concurrently for even better performance. Wrap batch operations in go routine to avoid blocking.
 
-### Batch Write Operations
+### (BatchWrite) Operations
 
 The following example demonstrates concurrent batch write operations:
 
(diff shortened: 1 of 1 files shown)


docs/async_api_usage.md
```diff
@@ -0,0 +1,766 @@
...
+- **Single Async Operations**: Running individual Put/Get operations
+- **Batch Async Operations**: Running multiple batch operations concurrently or leveraging built-in batch concurrency
+
+Unlike some other language clients, the Go client doesn't have explicit async/await patterns. Instead, Go's native concurrency primitives (goroutines and channels) are used to achieve asynchronous behavior.
+
+## Prerequisites
+
+Before using async operations, ensure you have:
+
+1. **Go 1.18+** installed
```
mirzakaracic: Issue: Match the version of go set in go.mod 

Go version in go.mod is set to `1.24.0`.

docs/async_api_usage.md
```diff
@@ -0,0 +1,766 @@
...
+	} else if result.Record == nil {
+		log.Printf("Key %s not found", result.Key.Value())
+	} else {
+		fmt.Printf("Successfully read key %s: %v\n", result.Key.Value(), result.Record.Bins)
+	}
+
+  
+```
+
+### Operate Operations
```
mirzakaracic: nit: format of the title 

`(Operate) Operation`

docs/async_api_usage.md
```diff
@@ -0,0 +1,766 @@
...
+
+	if result.Error != nil {
+		log.Printf("Error writing key %s: %v", result.Key.Value(), result.Error)
+	} else {
+		fmt.Printf("Successfully wrote key %s\n", result.Key.Value())
+	}
+
+```
+
+### Read Operations (Get)
```
mirzakaracic: nit: format of the title 

`(Get) Operation`

docs/async_api_usage.md
```diff
@@ -0,0 +1,766 @@
...
+batchPolicy.ConcurrentNodes = 5  // Number of concurrent node operations
+```
+
+## Single Async Operations
+
+Single async operations let you execute individual Put/Get calls independently. By using goroutines, you can run these operations concurrently on different records. The example below demonstrates how to perform a Get/Put on a single record using a goroutine — and the same pattern can be extended to handle multiple records concurrently.
+
+**Note**: A single Get/Put call does not require a goroutine; it can be invoked directly. The provided examples use goroutines only to show how these operations can participate in concurrent workflows.
+
+### Write Operations (Put)
```
mirzakaracic: nit: format of the title 

`(Write) Operation`

docs/async_api_usage.md
```diff
@@ -0,0 +1,766 @@
...
+		log.Printf("No record returned for key %s", res.Key.Value())
+	} else {
+		fmt.Printf("Operate successful for key %s: %v\n", res.Key.Value(), res.Record.Bins)
+	}
+
+```
+
+## Batch Async Operations
+
+Batch operations in the Aerospike Go client already use internal concurrency (via goroutines) to process multiple nodes in parallel. However, you can run multiple batch operations concurrently for even better performance.
```
mirzakaracic: nit: reword 

Replace `However, you can run multiple batch operations concurrently for even better performance.` with `Wrap batch operations in go routine to avoid blocking`

docs/async_api_usage.md
mirzakaracic: Will omit this comment for other places.

mirzakaracic (changes requested):
I would give this document to Stever Renekar. The ticket was asking to create a doc on we portal not here. I think there is some confusion what I suggested would be nice to have for reviews.

dmishra-aero:
> I would give this document to Stever Renekar. The ticket was asking to create a doc on we portal not here. I think there is some confusion what I suggested would be nice to have for reviews.

Yes, the idea is to avoid adding these documentation files to the code repository. I’ve only placed the document in the docs directory for review, since that’s where the other documents are stored.

docs/async_api_usage.md
dmishra-aero: done

docs/async_api_usage.md
dmishra-aero: done

dmishra-aero:
Closing it. as the PR meant for Document sake

PR #564, opened 2025-12-08
CLIENT-3954 Instantiate DynConfig.logUpdate to avoid nil pointer dereference

reugn:
This PR ensures `DynConfig.logUpdate` is instantiated when dynamic configuration is enabled. This prevents a nil pointer dereference that occurs when attempting to access the uninitialized property.


mirzakaracic (approved):

mirzakaracic:
Thank you @reugn

PR #568, opened 2025-12-12
CLIENT-3939 index out of range panic occurs on 'newBatchOperateNodeListIfcRetry'

mirzakaracic:


mirzakaracic:
@khaf Did also validate by adding a random timeout error `ifc.getConnection(policy)` in command.go. The panic is not occurring with above fix.

khaf (approved):
Looks good.

PR #578, opened 2025-12-18
CLIENT-4009: field section was not parsed before reading the record

dmishra-aero:


mirzakaracic (dismissed):
👍 Looks good.

I have validated the change on local 3 node SC cluster. We have other failing tests we should fix, not part of this PR, to work on cluster in SC mode.

I would let @khaf also take a peek at this just to make sure.

aerospike_suite_test.go
```diff
@@ -183,8 +183,8 @@ func isEnterpriseEdition() bool {
 	node := client.GetNodes()[0]
 	serverVersion := node.GetServerVersion()
 	serverCommand := types.Ternary(
-		serverVersion.IsGreaterOrEqual(version.ServerVersion_8_1), 
-		"release", 
+		serverVersion.IsGreaterOrEqual(version.ServerVersion_8_1),
```
khaf: No such info key. Should be reverted.

khaf (changes requested):
An issue with the info key on the server should be addressed.

aerospike_suite_test.go
dmishra-aero: I don't see any difference with the earlier code ? Do you want me to remove this `version.ServerVersion_8_1` ?

dmishra-aero: Done

mirzakaracic: @khaf is referring to is `release` if you try to run this on server version 8.1.0,  `ServerVersion_8_1  = &Version{Major: 8, Minor: 1, Patch: 0, Build: 0}`, you will get an error. The release command is replacing `edition` command but this will be the case in 8.1.1. 

Having said that this should be 

```go
serverCommand := types.Ternary(
		serverVersion.IsGreater(version.ServerVersion_8_1), 
		"release", 
		"edition")
```
	
I am going to create a ticket to fix the case where we are performing the same check against server version 8.1.0 to get edition/release of server. I believe you got this from https://github.com/aerospike/aerospike-client-go/blob/174a824b25f0440cd94d24b64b875910d448dc09/aerospike_suite_test.go#L178

https://aerospike.atlassian.net/wiki/spaces/CLIENTS/pages/4616257553/Deprecated+Info+commands#Server-8.1.1

mirzakaracic: PR resolves above discovered issue. https://github.com/aerospike/aerospike-client-go/pull/584

dmishra-aero: @mirzakaracic Thanks for detailed explanation..This clarifies a lot of thing

khaf (approved):
Looks good.

PR #579, opened 2025-12-18
CLIENT-4016 BatchDelete panics in goroutine when namespace does not exist

mirzakaracic:
- Added logic to handle the case if node is nil for cases where we try to perform BatchDelete on record for namespace that does not exist.
- Will properly propagate error up the chain. This might be controversial change in that if there are errors on BatchDelete we are no longer returning nil but records. I would say border line api change. Not changing the API per se but the api behaves slightly differently. On the flip side this change does fall in line how Java functions.

Will have to back port to v6(confirmed exists) and v7.


khaf (approved):
Looks good.

PR #583, opened 2025-12-22
Client 3898 go cov pipeline

dmishra-aero:
1. Added a mechanism to build up a single node cluster with SC Mode 
2. Added a custom action to build up a single-node EE cluster setup especially for SC set up
3. Added a mechanism to build up a multiNode cluster with SC Mode
4. Added support for prompting user to provides modes in workflow UI
5. Created a re-usable `merge coverage` workflow generating merged coverage report across all job happened within a workflow
6. Created a secret in repo containing base64 encoded content of **_feature.conf_** file used by Multi node set-up jobs
7. Skipped a failing test case for now which is tracked here -> CLIENT-4009
8. Added a custom config file that has SC enabled
9. For probing the coverage report for specific or overall files , user can either download the `coverage-html ` folder from the uploaded artifact and can do following things
    -    open the` index.html `file to see the uncovered lines

**Note :**
- `reugn/github-action-aerospike@v1` is unusable for single-node EE cluster setup because of bugs.


.github/actions/deploy-single-node-sc-cluster/wait-for-as-server-to-start.bash
```diff
@@ -0,0 +1,47 @@
...
+set -o pipefail
+
+# We use bash because we need the not (!) operator
+
+call_from_tools_container() {
+    # See comment in run-ee-server.bash's call_from_tools_container() for why we surround $@ with double quotes
+    docker run --rm --network host aerospike/aerospike-tools "$@"
+}
+
+while true; do
```
khaf: This will hang indefinitely if the server does not start for some reason.

khaf: Some kind of timeout mechanism should be implemented here, maybe for 30s or other acceptable duration.

.github/workflows/build-multi-node.yml
```diff
@@ -102,12 +111,40 @@
...
+
+  merged_coverage:
+    name: Final Coverage
+    needs: build
+    uses: ./.github/workflows/merge-coverage.yml
+    with:
+      artifact_pattern: cover_*
+      threshold: 65
+
+
```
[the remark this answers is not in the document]
khaf: Probably should be addressed.

.github/workflows/build.yml
khaf, line 43: here as well.

.github/workflows/build.yml
```diff
@@ -102,12 +109,40 @@ jobs:
...
+          path: reports
+          if-no-files-found: warn
+
+  merged_coverage:
+    name: Final Coverage
+    needs: build
+    uses: ./.github/workflows/merge-coverage.yml
+    with:
+      artifact_pattern: cover_*
+      threshold: 65
```
khaf: Same here.

.github/workflows/merge-coverage.yml
```diff
@@ -0,0 +1,66 @@
...
+          mkdir -p coverage-reports
+          go tool cover -html=cover_all.out -o coverage-reports/index.html
+          mv cover_all.out coverage-reports/
+
+      - name: Upload Html report
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage-html
+          path: coverage-reports
+          if-no-files-found: error
```
khaf, line 70: And here.

aerospike_suite_test.go
```diff
@@ -183,8 +202,8 @@ func isEnterpriseEdition() bool {
 	node := client.GetNodes()[0]
 	serverVersion := node.GetServerVersion()
 	serverCommand := types.Ternary(
-		serverVersion.IsGreaterOrEqual(version.ServerVersion_8_1), 
-		"release", 
+		serverVersion.IsGreaterOrEqual(version.ServerVersion_8_1),
```
khaf, line 205: No such key available as of `v8.1.0.1`. Should be reverted.

khaf (changes requested):
There are some issues in the workflow and test infra which should be addressed.

.github/actions/deploy-single-node-sc-cluster/wait-for-as-server-to-start.bash
dmishra-aero: I guess git step have some default timeout if such thing happen , 
but it make sense to have our own timeout, we can configure at git `step` level which would be more cleaner

mirzakaracic: I would recommend timeout with exponential and multiplied delay on every iteration. Cap at certain number of intervals. I would make timeout and number of intervals configurable via actions vars.

.github/config/aerospike-sc.conf
```diff
@@ -0,0 +1,68 @@
+# Aerospike database configuration file
```
mirzakaracic: To enable sc mode all you have to do is `aerolab conf sc`. In fact anything you need to change with the conf you should be able to do with `aerolab conf....`

❯ al conf
Usage:
  aerolab [OPTIONS] conf [command]

Global Options:
  --beep     cause the terminal to beep on exit; if specificied multiple times, will be once on success and >1 on failure
  --beepf    like beep, but does not trigger beep on success, only failures

Available commands:
  generate          Generate or modify Aerospike configuration files
  sc                Configure the cluster to use strong-consistency, with roster and optional RF changes
  fix-mesh          Fix mesh configuration in the cluster
  rackid            Change/add rack-id to namespaces in the existing cluster nodes
  namespace-memory  Adjust memory for a namespace using total percentages
  adjust            Adjust running Aerospike configuration parameters
  help              Print help

Unless `aerolab conf...` does not allow you to set an option I would not create configurations file.

.github/workflows/build.yml
```diff
@@ -9,7 +11,7 @@ env:
   GOFIPS140: latest
 jobs:
   make-matrix:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-latest
```
mirzakaracic: I would not use lates version of anything. Things start breaking for the moment new disto is out. I would pin to a version. Also make this configurable via vars. Take a look at this line 

https://github.com/aerospike/aerospike-client-java/blob/47fb080784a7a6205d9c92b0a6f98f4e041d23a4/.github/workflows/promote.yaml#L35

I would make sure to fix this in all the workflows you have added/updated.

aerospike_suite_test.go
dmishra-aero: PR has already been created here -https://github.com/aerospike/aerospike-client-go/pull/584

.github/config/aerospike-sc.conf
dmishra-aero: I tried creating a cluster using Aerolab and then running the `aerolab conf sc `command, but it didn’t work. Although the configuration appears to be updated when inspecting the node, the cluster does not actually operate in SC mode, and the test cases fail. However, when I create the cluster by providing a custom configuration upfront, the cluster operates correctly in SC mode, and all the test cases pass. Even with aerolab if you use `aerolab cluster create -v 8.1.0.1 -n aero-sc -o /.github/config/aerospike-sc.conf` then this works but we need to pass config file upfront

dmishra-aero: Additionally, the configuration requires the flag `strong-consistency-allow-expunge true` without it, some test cases particularly delete operation fail with the error `Operation not allowed this time`. This setting, when enabled after cluster creation, does not take effect as expected. It only works correctly when supplied upfront as part of config file during cluster creation.


```
 default_json: |
            {
              "stable": { "version": "8.1.0.1" },
              "release": { "version": "8.1.0.1_2" }
        }
```
  Also if we use aerolab to download above version during single node deployment, the `release` one will fail as they are docker tags..but not present in `https://download.aerospike.com/artifacts/aerospike-server-enterprise/ `which aerolab uses to fetch the artifact from

.github/workflows/build.yml
dmishra-aero: Done

.github/workflows/merge-coverage.yml
dmishra-aero: Done

.github/workflows/build.yml
dmishra-aero: Done

.github/actions/deploy-single-node-sc-cluster/wait-for-as-server-to-start.bash
dmishra-aero: Done

.github/workflows/build-multi-node.yml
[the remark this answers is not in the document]
dmishra-aero: done

khaf (approved):
Looks good to me. I'll let @mirzakaracic to give the final thumbs up and merge though.

mirzakaracic (approved):

PR #584, opened 2025-12-22
CLIENT-4031  command usage check not done against correct version of AES server

mirzakaracic:


mirzakaracic:
Fixed in other PR's

PR #597, opened 2026-01-16
CLIENT-4067 Support Privilege / Permission Code Expansion Due to DataMasking Feature

mirzakaracic:


admin_command.go
```diff
@@ -676,7 +676,12 @@ func (acmd *AdminCommand) parseRolesFull(receiveSize int) (int, []*Role, Error)
 				role.Name = string(acmd.dataBuffer[acmd.dataOffset : acmd.dataOffset+len])
 				acmd.dataOffset += len
 			case _PRIVILEGES:
+				startOffset := acmd.dataOffset
```
khaf: This is an anti-pattern. If `parsePrivileges` has an issue, it should be fixed in that method. Just skipping bytes is also not a good pratice, we should know what they are.

privilege.go
```diff
@@ -81,9 +83,28 @@ func (p *Privilege) code() int {
...
+		return -1
+
+	default:
+		// This should never happen if privilegeFrom() is used correctly
+		logger.Logger.Warn("Invalid privilege code in Privilege.code(): %s", p.Code)
+		return -1
+	}
 }
 
 func privilegeFrom(code uint8) privilegeCode {
```
khaf: This method should be modified to return an error, and that error should be propagated back to the user.

privilege.go
```diff
@@ -136,11 +157,31 @@ func privilegeFrom(code uint8) privilegeCode {
...
+	// User can write data with masking policies applied. Requires server version >= 8.1.1.
+	case 17:
+		return WriteMasked
+
+	default:
+		// Return Unknown for forward compatibility with new privilege codes
+		// from future server versions. This allows the client to work with
+		// newer servers without breaking when new privileges are added.
+		logger.Logger.Warn("Unknown privilege code received from server: %d. Using Unknown.", code)
+		return Unknown
```
khaf: Didn't we decide not to do this? This only makes sense for the C client.

role.go
```diff
@@ -64,4 +64,18 @@ const (
...
+	// MaskingAdmin allows to manage masking policies. Requires server version >= 8.1.1.
+	MaskingAdmin privilegeCode = "masking-admin"
+
+	// ReadMasked allows read transactions with the database with masking policies applied. Requires server version >= 8.1.1.
+	ReadMasked privilegeCode = "read-masked"
+
+	// WriteMasked allows write transactions with the database with masking policies applied. Requires server version >= 8.1.1.
+	WriteMasked privilegeCode = "write-masked"
+
+	// Unknown privilege code from server.
```
khaf: Not needed. An error should be returned to the user.

khaf (changes requested):
Too many design and implementation issues. Further discussions required.

role.go
```diff
@@ -64,4 +64,18 @@ const (
...
+	// This is used for forward compatibility when the server sends
+	// privilege codes that are not yet known to this client version.
+	Unknown privilegeCode = "unknown"
```
khaf: Should not be exported.

admin_command.go
```diff
@@ -306,29 +306,33 @@ func (acmd *AdminCommand) writePrivileges(privileges []Privilege) Error {
...
-		offset++
-
-		if privilege.canScope() {
-
-			if len(privilege.SetName) > 0 && len(privilege.Namespace) == 0 {
-				return newError(types.INVALID_PRIVILEGE, fmt.Sprintf("Admin privilege '%v' has a set scope with an empty namespace.", privilege))
-			}
-
-			acmd.dataBuffer[offset] = byte(len(privilege.Namespace))
+		if code := privilege.code(); code < 0 {
```
khaf: If there is a validation on the privilege, it should be moved to a method on the type itself.

khaf: Long if/else block is also not a good practice. Just verify first and exit, otherwise continue.

admin_command.go
```diff
@@ -664,7 +668,7 @@ func (acmd *AdminCommand) parseRolesFull(receiveSize int) (int, []*Role, Error)
 		fieldCount := int(acmd.dataBuffer[acmd.dataOffset+3])
 		acmd.dataOffset += _HEADER_REMAINING
 
-		for i := 0; i < fieldCount; i++ {
+		for range fieldCount {
```
khaf, line 669: The library uses Go 1.21, while this range syntax was introduced in Go v1.22. Need to be reverted until the next major release.

admin_command.go
```diff
@@ -702,17 +708,22 @@ func (acmd *AdminCommand) parseRolesFull(receiveSize int) (int, []*Role, Error)
...
-func (acmd *AdminCommand) parsePrivileges(role *Role) {
+func (acmd *AdminCommand) parsePrivileges(role *Role) Error {
 	size := int(acmd.dataBuffer[acmd.dataOffset] & 0xFF)
 	acmd.dataOffset++
 	role.Privileges = make([]Privilege, 0, size)
 
-	for i := 0; i < size; i++ {
-		priv := Privilege{}
-		priv.Code = privilegeFrom(acmd.dataBuffer[acmd.dataOffset])
+	for range size {
```
khaf, line 714: Same here.

admin_command.go
```diff
@@ -306,29 +306,33 @@ func (acmd *AdminCommand) writePrivileges(privileges []Privilege) Error {
...
-			offset += len(privilege.Namespace)
 
-			acmd.dataBuffer[offset] = byte(len(privilege.SetName))
-			offset++
-			copy(acmd.dataBuffer[offset:], privilege.SetName)
-			offset += len(privilege.SetName)
-		} else {
-			if len(privilege.Namespace) > 0 || len(privilege.SetName) > 0 {
-				return newError(types.INVALID_PRIVILEGE, fmt.Sprintf("Admin global rivilege '%v' can't have a namespace or set.", privilege))
+			if canScope, err := privilege.canScope(); err != nil {
```
khaf: Long if/else block.

privilege.go
```diff
@@ -81,66 +85,92 @@ func (p *Privilege) code() int {
...
+		return WriteMasked, nil
+
+	default:
+		return Unknown, newError(types.INVALID_PRIVILEGE, fmt.Sprintf("Unknown privilege code received from server: %d.", code))
+	}
 }
 
-func (p *Privilege) canScope() bool {
-	return p.code() >= 10
+func (p *Privilege) canScope() (bool, Error) {
```
khaf: Error here is useless. Should be removed.

khaf (changes requested):
I think there was way too much change due to the unnecessary error introduced on the `canScope()` method. Removing it will significantly simplify the changeset.

admin_command.go
mirzakaracic: I think we have a problem if that is the case. Stage is on 1.24. It looks like I have cause that issue when I merged dependabot PR. However v8 is on 1.23.0. It looks like we have bumped the version some time ago as part of snyk security reports. We should chat about this one.

khaf (approved):
Looks good.

PR #603, opened 2026-01-22
CLIENT-4093 Update go to use any{} instead of interface{}

mirzakaracic:


khaf (approved):
I can't possibly go through all of this changeset, but at a cursory glance it looks good.

PR #604, opened 2026-01-23
CLIENT-4083:Added sleepMutliplier to be part of dynamic config

dmishra-aero:
Added `sleepMutliplier` param to be part of dynamic config,
Also ensure `sleepMutliplier` to be applied consistently in conjunction with `sleepBetweenRetries`

### DIFF
--- a/config/dynconfig.go
+++ b/config/dynconfig.go
@@ -86,6 +86,7 @@ type Read struct {
 	ReadModeSc          *ReadModeSc `yaml:"read_mode_sc"`
 	Replica             *Replica    `yaml:"replica"`
 	SleepBetweenRetries *int        `yaml:"sleep_between_retries"`
+	SleepMultiplier     *float64    `yaml:sleep_multiplier`
 	SocketTimeout       *int        `yaml:"socket_timeout"`
 	TotalTimeout        *int        `yaml:"total_timeout"`
 	MaxRetries          *int        `yaml:"max_retries"`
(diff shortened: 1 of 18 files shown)


mirzakaracic:
@dmishra-aero Is this still WIP? Tests are failing.

dmishra-aero:
@mirzakaracic I used a hardcoded value for local testing and missed removing it before pushing.

config/dynconfig.go
```diff
@@ -86,6 +86,7 @@ type Read struct {
 	ReadModeSc          *ReadModeSc `yaml:"read_mode_sc"`
 	Replica             *Replica    `yaml:"replica"`
 	SleepBetweenRetries *int        `yaml:"sleep_between_retries"`
+	SleepMultiplier     *float64    `yaml:sleep_multiplier`
```
mirzakaracic, line 89: question: Why is SleepMultiplier a float?

dmishra-aero: This was already defined as a `float64` in `policy.go` before this change, so keeping it the same avoids unnecessary churn. Using `float64` for a backoff multiplier is also pretty standard (for example, [github.com/cenkalti/backoff](https://pkg.go.dev/github.com/cenkalti/backoff/v5?utm_source=chatgpt.com#section-readme) does the same).

mirzakaracic (approved):

khaf (approved):
Looks good.

PR #605, opened 2026-01-26
CLIENT-3968 Performance regression with in hot path with detailed metrics

mirzakaracic:
Summary of changes made so far
- using predefined size of array for commands in metrics collection
- instead of using interface call for getNamespace and or getNamespaces will be passing in namespace in the updateOrInsert
- removed lock on updateOrInsert

## Results 

BenchmarkUpdateOrInsertHighConcurrency

  OLD (Jan 23):  249,214 ns/op    79,261 B/op    3,201 allocs/op    (48,040 iters)
  NEW (Jan 26):   38,728 ns/op    63,218 B/op    2,201 allocs/op   (275,412 iters)
  IMPROVEMENT:    6.4x faster      20% less mem   31% fewer allocs

 BenchmarkUpdateOrInsertHighConcurrencyContended

  OLD (Jan 23):  22,831,134 ns/op    24,812,227 B/op    400,223 allocs/op    (524 iters)
  NEW (Jan 26):   1,403,183 ns/op     5,610,614 B/op    200,203 allocs/op  (8,041 iters)
  IMPROVEMENT:    16.3x faster        77% less mem      50% fewer allocs

 BenchmarkUpdateOrInsertScalability

  goroutines_1:
    OLD:    499.5 ns/op      352 B/op     7 allocs/op
    NEW:    425.3 ns/op      160 B/op     5 allocs/op
            1.2x faster      54% less     29% fewer

  goroutines_10:
    OLD:   4,796 ns/op     3,376 B/op    61 allocs/op
    NEW:   3,204 ns/op     1,456 B/op    41 allocs/op
           1.5x faster     57% less      33% fewer

  goroutines_50:
    OLD:  24,502 ns/op    16,819 B/op   301 allocs/op
    NEW:  12,856 ns/op     7,216 B/op   201 allocs/op
          1.9x faster      57% less     33% fewer

  goroutines_100:
    OLD:  51,656 ns/op    33,623 B/op   601 allocs/op
    NEW:  24,861 ns/op    14,416 B/op   401 allocs/op
          2.1x faster      57% less     33% fewer

  goroutines_200:
    OLD: 101,273 ns/op    67,235 B/op   1,201 allocs/op
    NEW:  51,219 ns/op    28,817 B/op     801 allocs/op
          2.0x faster      57% less      33% fewer

  goroutines_500:
    OLD: 251,719 ns/op   168,095 B/op   3,001 allocs/op
    NEW: 129,325 ns/op    72,019 B/op   2,001 allocs/op
          1.9x faster      57% less      33% fewer

SUMMARY

  High contention:   16x faster, 77% memory reduction
  High concurrency:  6x faster, 31% fewer allocations
  Scalability:       ~2x speedup at high goroutine counts, ~57% memory reduction


run_updateorinsert_benchmark.sh
```diff
@@ -0,0 +1,109 @@
...
+    local bench_name=$1
+    local description=$2
+    
+    echo "Running: $bench_name"
+    echo "Description: $description"
+    echo ""
+    
+    # Run benchmark with CPU, memory, and mutex profiling
+    # -run='^$' skips all regular tests and runs only benchmarks
+    go test -run='^$' -bench="^${bench_name}$" \
```
khaf: Probably need to pass `-r` to it to make sure we don't hit a race condition.

batch_command_delete.go
```diff
@@ -119,7 +119,7 @@ func (cmd *batchCommandDelete) parseRecordResults(ifc command, receiveSize int)
 		// Aggregate metrics
 		metricsEnabled := cmd.node.cluster.metricsEnabled.Load()
 		if metricsEnabled {
-			cmd.node.stats.updateOrInsert(ifc, resultCode)
+			cmd.node.stats.updateOrInsert(ifc.getNamespace(), ifc.getNamespaces(), ifc.commandType(), resultCode)
```
khaf: `ifc.*` is a virtual call. On the object itself, we can call `cmd.*` for a static call (Except on some specific commands like `StreamCommand`).

khaf (changes requested):
Looks good. Is this change backwards compatible with what the user used to get from the `client.GetStats` call?

batch_command_delete.go
khaf: Here I meant everywhere :)

khaf (changes requested):
All the `ifc.XXX()` method calls need to be converted to `cmd.XXX()`

khaf (approved):
Looks good.

PR #615, opened 2026-02-03
CLIENT-4155 Invalid binNames on selectByPath and modifyByPath return parameter error

mirzakaracic:


bin.go
```diff
@@ -14,12 +14,17 @@
 
 package aerospike
 
+const (
+	// MaxBinNameLength is the maximum bin name length.
+	MaxBinNameLength = 15
```
khaf: Should this be exposed to the user?

bin.go
```diff
@@ -14,12 +14,17 @@
...
+)
+
 // BinMap is used to define a map of bin names to values.
 type BinMap map[string]any
 
 // Bin encapsulates a field name/value pair.
 type Bin struct {
-	// Bin name. Current limit is 14 characters.
+	// Bin name. Current limit is 15 characters.
```
khaf: Bytes. Runes (Unicode characters) are multi-byte in Go.

operation.go
```diff
@@ -64,6 +64,9 @@ type Operation struct {
 
 	// binValue (Optional) determines bin value used in operation.
 	binValue Value
+
+	// err (Optional) contains validation error from operation creation.
+	err Error
```
khaf: This should be removed now.

command.go
```diff
@@ -4213,3 +4225,8 @@ func (cmd *baseCommand) applyDetailedMetricsDataSizeAndLatencyOnWrite(ifc comman
 		}
 	}
 }
+
+func (cmd *baseCommand) validateBinName(binName string) (int, bool) {
```
khaf: The name should reflect what the method does. The better names would probably be `writeAndValidateBinName`.

khaf (changes requested):

bin.go
mirzakaracic: Not it should not.

khaf (approved):
Looks good.

exherb:
why empty bin name is not allowed and these kind of changes are not mark as breaking change? @mirzakaracic @khaf

PR #619, opened 2026-02-13
CLIENT-3885 Add a guardrail to prevent expiration from being set with  read-only operate

mirzakaracic:
Added test to validate since the logic was already in place.


khaf (approved):
Looks good.

PR #640, opened 2026-03-25
CLIENT-3998 Ops Projection

mirzakaracic:


khaf (dismissed):
Looks good.

internal/version/version.go
```diff
@@ -30,6 +30,7 @@ type Version struct {
 
 var (
 	ServerVersion_8_1           = &Version{Major: 8, Minor: 1, Patch: 0, Build: 0}
+	ServerVersion_8_1_2         = &Version{Major: 8, Minor: 1, Patch: 2, Build: 0}
```
khaf: One of the better ways to do this would be to add `version.SupportsXXX()` methods to remove the ambiguity on the call sites.

node_validator.go
```diff
@@ -322,6 +322,9 @@ func (ndv *nodeValidator) setFeatures(alias *Host) Error {
 		ndv.features |= _SUPPORTS_BATCH_ANY
 		ndv.features |= _SUPPORTS_PARTITION_QUERY
 	}
+	if ndv.serverVersion.IsGreaterOrEqual(version.ServerVersion_8_1_2) {
```
khaf: So here, the call would be:

```go
if version.SupportsQueryOpsProjectionExt() {
...
}
```

khaf: The current design is a leaky abstraction. Not a huge deal, but the alternative is nicer and easier to read.

khaf (dismissed):
Looks OK. I have added a comment regarding a change to the `Version` struct. I have approved the PR, but won't merge it. You can merge it yourself based on your own discretion.

PR #639, opened 2026-03-20
CLIENT-4564 Fix batch executeSingle ignores errors when AllowPartialResults is true

matthewd98:
Relates to https://github.com/aerospike/aerospike-client-go/issues/638

PR #650, opened 2026-04-02
CLIENT-4564 Fix batch executeSingle ignores errors when AllowPartialResults is true (v8)

matthewd98:
Branched off of stage

Original PR: https://github.com/aerospike/aerospike-client-go/pull/639


matthewd98:
@mirzakaracic Looks like there are many other unmerged changes in the stage branch. I'll wait for your guidance on how you'd like the PR for v8 prepared -- whether you prefer that I merge my changes into the stage branch or follow another path.

PR #655, opened 2026-04-08
CLIENT-4632 Fixes executeSingle iterates all records instead of node-assigned offsets

vahid-sohrabloo:
When a batch node has exactly 1 key assigned to it, several batch command types fall back to `executeSingle` which
   incorrectly iterates **all records/keys** in the entire batch instead of only the record(s) assigned to the current node
    via `cmd.batch.offsets`.

   This causes N sequential `client.Operate` round trips instead of 1, where N is the total number of keys in the batch.

   ### Impact

   On cross-datacenter connections with ~2.5ms RTT, a `BatchGetComplex` with 10 keys takes **~25ms instead of ~3ms**
   because `executeSingle` sends 10 individual Operate calls sequentially.

   ### Root cause

   Compare the **correct** implementation in `batchCommandGet.executeSingle`:

   ```go
   // Correct: iterates only this node's assigned keys
   for _, offset := range cmd.batch.offsets {
       cmd.records[offset], err = client.Get(...)
   }
   ```

   With the **buggy** implementation in `batchIndexCommandGet.executeSingle`:

   ```go
   // Bug: iterates ALL records in the batch
   for _, br := range cmd.records {
       client.Operate(...)
   }
   ```

   ### Affected files

   | File | Execute condition | Loop iterates | Bug |
   |------|------------------|---------------|-----|
   | `batch_command_get.go` | `len(cmd.batch.offsets) == 1` | `cmd.batch.offsets` | ✅ Correct |
   | `batch_command_exists.go` | `len(cmd.batch.offsets) == 1` | `cmd.batch.offsets` | ✅ Correct |
   | `batch_index_command_get.go` | `len(cmd.batch.offsets) == 1` | `cmd.records` | ❌ Bug |
   | `batch_command_operate.go` | `len(cmd.records) == 1` | `cmd.records` | ❌ Bug |
   | `batch_command_delete.go` | `len(cmd.keys) == 1` | `cmd.keys` | ❌ Bug |
   | `batch_command_udf.go` | `len(cmd.keys) == 1` | `cmd.keys` | ❌ Bug |

   ### Fix

   For all affected files:
   1. Check `len(cmd.batch.offsets) == 1` instead of `len(cmd.records/keys) == 1`
   2. Iterate `cmd.batch.offsets` and index into `cmd.records/keys` by offset

   ### Reproduction

   Test against a 3-node cluster with metrics enabled:

   ```go
   client.EnableMetrics(aero.DefaultMetricsPolicy())
   _, _ = client.Stats() // reset

   for range 100 {
       reads := make([]*aero.BatchRead, 10)
       for i := range 10 {
           reads[i] = aero.NewBatchRead(nil, keys[i], []string{"val"})
       }
       client.BatchGetComplex(bp, reads)
   }

   stats, _ := client.Stats()
   // Check operate-metrics count in cluster-aggregated-stats
   // Expected: 0 (all should use batch protocol)
   // Actual: 1000 (10 Operate calls per iteration)

### DIFF
--- a/batch_command_delete.go
+++ b/batch_command_delete.go
@@ -186,11 +186,12 @@ func (cmd *batchCommandDelete) commandType() commandType {
 
 func (cmd *batchCommandDelete) executeSingle(client *Client) Error {
 	policy := cmd.batchDeletePolicy.toWritePolicy(cmd.policy, client.dynConfig)
-	for i, key := range cmd.keys {
+	for _, offset := range cmd.batch.offsets {
+		key := cmd.keys[offset]
 		res, err := client.Operate(policy, key, DeleteOp())
-		cmd.records[i].setRecord(res)
+		cmd.records[offset].setRecord(res)
 		if err != nil {
-			cmd.records[i].setRawError(err)
+			cmd.records[offset].setRawError(err)
 
 			// Key not found is NOT an error for batch requests
 			if err.resultCode() == types.KEY_NOT_FOUND_ERROR {
@@ -209,7 +210,7 @@ func (cmd *batchCommandDelete) executeSingle(client *Client) Error {
 }
 
 func (cmd *batchCommandDelete) Execute() Error {
-	if len(cmd.keys) == 1 {
+	if len(cmd.batch.offsets) == 1 {
 		return cmd.executeSingle(cmd.client)
 	}
 	return cmd.execute(cmd)
--- a/batch_command_operate.go
+++ b/batch_command_operate.go
@@ -239,7 +239,8 @@ func (cmd *batchCommandOperate) parseRecord(key *Key, opCount int, generation, e
 func (cmd *batchCommandOperate) executeSingle(client *Client) Error {
 	var res *Record
 	var err Error
-	for _, br := range cmd.records {
+	for _, offset := range cmd.batch.offsets {
+		br := cmd.records[offset]
 
 		switch br := br.(type) {
 		case *BatchRead:
@@ -288,7 +289,7 @@ func (cmd *batchCommandOperate) executeSingle(client *Client) Error {
 }
 
 func (cmd *batchCommandOperate) Execute() Error {
-	if cmd.objects == nil && len(cmd.records) == 1 {
+	if cmd.objects == nil && len(cmd.batch.offsets) == 1 {
 		return cmd.executeSingle(cmd.client)
 	}
 	return cmd.execute(cmd)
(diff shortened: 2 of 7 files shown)


vahid-sohrabloo:
@khaf  when have time can you check this?

mirzakaracic:
👀

mirzakaracic (commented):
So this does seem to be an issue just looking at the java and C clients. For instance the patterns seems to be:

```
  if (bn.offsetsSize == 1) {
      int i = bn.offsets[0];     // extract the ONE assigned offset
      // pass keys[i] or records[i] to BatchSingle command
  }
```
https://github.com/aerospike/aerospike-client-java/blob/91c67d1865f0b734d19ad85aa6d00c10b12c8f6b/client/src/com/aerospike/client/AerospikeClient.java#L1265
https://github.com/aerospike/aerospike-client-java/blob/91c67d1865f0b734d19ad85aa6d00c10b12c8f6b/client/src/com/aerospike/client/AerospikeClient.java#L1417

Having said that this PR is not complete since it's missing https://github.com/aerospike/aerospike-client-go/commit/a91785d15db052e3929c60f56b8d5300656fcdaa. I have pushed the changes https://github.com/aerospike/aerospike-client-go/tree/execute-single-iterates. The correnct extent of the changes should be https://github.com/aerospike/aerospike-client-go/compare/stage...execute-single-iterates.

I will let @khaf chime in just to make sure I am not missing something here.

khaf:
Sorry for my late reply. The fix makes sense, but we need tests that would trigger the issue and capture the fix.  Can you submit them please @vahid-sohrabloo ?

vahid-sohrabloo:
@khaf, as I understand, @mirzakaracic  is moving to another branch and adding some tests. Not sure how I can help.

mirzakaracic:
@vahid-sohrabloo I am going to update the branch `execute-single-iterates` and add more tests. Should be able to get that done some time today.

mirzakaracic:
@vahid-sohrabloo I would like you to give you credit since you have found and fixed the issue. To create a viable PR you will have to 

1. Fork a new branch off of stage.
2. Pull in the changes from `execute-single-iterates`
3. Force push to current PR or create another.
4. Make sure the base branch the PR is for is set to stage not v8.

Let me know if you would like to go this route otherwise I will create a new PR.

vahid-sohrabloo:
@mirzakaracic Thanks. I did it.

mirzakaracic:
@vahid-sohrabloo Thank you!

khaf (approved):
Looks good.

mirzakaracic (approved):

PR #667, opened 2026-04-21
CLIENT-4688 Add inline annotation

mirzakaracic:


mirzakaracic:
Met with @khaf . Came to following conclusion

- Will reach out to customer to find out exactly what they are doing/how they are planning to use this feature
- Once we have that information we are going to look into weather this needs to be done in the first place since go has made optimization over the years which might render changes moot.
- Will create a new PR only if we have concrete yes on the first two bullet points.

PR #678, opened 2026-05-08
perf: restore multi-million ops/sec by reducing metrics hot-path overhead

### LINKED ISSUE #1: Optimisation computeDigest() in key.go?
I notice that the key.go: every time computing a digest for a key, it has to allocate a new slice. This will stress GC. In aerospike java client source code, I notice that thread-local buffer technique is used to improve this. But in Golang, there is no equivalent technique. I think why not to use a singleton buffer to hold data when computing key digest, instead of allocating a new slice many times?

icellan:
## Summary

The metrics / observability instrumentation added between v8.2.0 and v8.7.0 (CLIENT-2702, CLIENT-3439, CLIENT-3968) introduced enough per-operation overhead to substantially regress throughput on high-concurrency workloads — even when metrics are disabled. We measured the regression at multi-million ops/sec on our production workload and tracked it down to three independent issues. This PR fixes all three with small, surgical changes that preserve the existing public API and metrics-ON behaviour.

## The problems

1. **`internal/atomic.Int` is mutex-backed, not lock-free.**
   Every increment/decrement takes `sync.Mutex.Lock()`. There are ~70 hot-path usages on the per-command path: `TransactionRetryCount`, `TransactionErrorCount`, `ConnectionsRecovered`, every connection-pool counter, etc. These fire **regardless of whether metrics are enabled**, because they're not gated. On any workload with retries or transient errors the mutex contention dominates.

2. **The `metricsEnabled` gate is `atomic.Bool`, advertising runtime mutability we don't actually use.**
   `atomic.Bool.Load()` is cheap on amd64/arm64 (~1 ns), but it prevents the compiler from hoisting the read out of inner loops. A field that never changes after startup should be a plain `bool` — same nanoseconds, but the compiler can do more with it and the branch predictor sees a constant.

3. **Three helpers in `command.go` write metrics state unconditionally.**
   `applyTransactionErrorMetrics`, `applyTransactionRetryMetrics`, `applyConnectionRecoveredMetrics` all increment counters on every call without checking the gate. Even after fix #1 makes the increment cheap, the metrics-OFF path was paying for work that should have been zero-cost.

### 1. `internal/atomic/int.go` — lock-free implementation

Replaced `sync.Mutex + int` with `sync/atomic.AddInt64` etc. on a plain `int64` field. Public API is unchanged — every existing caller compiles and works as before. `go vet -copylocks` stays clean (`Clone` / `CloneAndSet` return the wrapper struct by value, which is safe with a plain `int64` field rather than `atomic.Int64` which has a `noCopy` tag).

### 2. `Cluster.metricsEnabled` / `Cluster.metricsPolicy` — plain fields

Changed from `atomic.Bool` / `iatomic.TypedVal[*MetricsPolicy]` to plain `bool` / `*MetricsPolicy`. Configured once at `NewCluster` from the new `ClientPolicy.MetricsPolicy` field. Reading the gate on the per-command hot path becomes a single non-synchronizing load.

The existing `Cluster.EnableMetrics` / `Cluster.DisableMetrics` and `Client.EnableMetrics` / `Client.DisableMetrics` methods are **retained for backwards compatibility**, with a docstring noting that calling them concurrently with in-flight commands is racy by design — set the policy once at startup and don't toggle. The dynamic-config callback in `metrics_policy.go` is unchanged and continues to work via these methods.

The 26 hot-path call sites that read the gate (across `command.go` and the per-command files) drop their `.Load()` suffix — purely mechanical.

### 3. Gate the unconditional helpers

`applyTransactionErrorMetrics`, `applyTransactionRetryMetrics`, `applyConnectionRecoveredMetrics` now short-circuit when the gate is off, so the metrics-OFF path skips the increment entirely.

## API impact

- **No public API removed.** `Client.EnableMetrics`, `Client.DisableMetrics`, `Cluster.EnableMetrics`, `Cluster.DisableMetrics`, `MetricsEnabled()`, `MetricsPolicy()` all retain their existing signatures and continue to function.
- **One new public field:** `ClientPolicy.MetricsPolicy *MetricsPolicy`. Nil (the default) means metrics off; non-nil enables metrics for the cluster lifetime, replacing the need to call `EnableMetrics` after construction.
- **Concurrency contract change** (documented in code): the cluster-level metrics fields are now plain (non-atomic). The supported pattern is to set `ClientPolicy.MetricsPolicy` once at startup. Calling `EnableMetrics` / `DisableMetrics` after commands are flowing is a data race and is documented as unsafe. Programs that already only call these methods at startup (the common pattern) are unaffected.

## Out of scope

The histogram lock contention on the metrics-ON path (`SyncHistogram.Add` takes a write-lock per call, no per-CPU sharding) is intentionally left for a separate change. It only matters when metrics are enabled, and a real fix there involves either sharded histograms or atomic-bucket counters — a larger discussion that doesn't belong in a \"fix the regression\" PR.

The connection-pool heap mutex (`connection_heap.go`) is also out of scope — it predates v8.2.0 and is unrelated to this regression.

## Verification

- [x] `go build ./...` clean.
- [x] `go vet .` warning count drops from 44 to 39 (five `atomic.Bool` / `noCopy` warnings disappear with the field-type change; zero new warnings introduced).
- [x] `go test ./internal/atomic/...` passes under `-race`.
- [x] `go test ./types/histogram/...` passes.
- [x] Spot-checked: `grep -rn 'metricsEnabled\.Load()' --include='*.go' .` returns zero matches.
- [x] Spot-checked: `EnableMetrics` / `DisableMetrics` still resolve on both `Client` and `Cluster`.
- [ ] **Throughput benchmark against a live cluster** — recommend running your standard benchmark against a v8.2.0 baseline before merging. We measured the recovery internally; happy to share methodology.

## Test plan

- [ ] Run the existing test suite against a live Aerospike cluster — same set of tests should pass as on `v8` (the offline tests fail identically with `INVALID_NAMESPACE` on this commit and on the `v8.7.0` baseline, confirming no regression).
- [ ] Run `go test -race ./...` against a live cluster.
- [ ] Run a single-client multi-goroutine `Get`/`Put` throughput benchmark with metrics OFF and confirm ops/sec is back to (or close to) v8.2.0 levels.
- [ ] Run the same with metrics ON and confirm counters and histograms still populate correctly via `Client.Stats()`.

### DIFF
--- a/batch_command_delete.go
+++ b/batch_command_delete.go
@@ -117,7 +117,7 @@ func (cmd *batchCommandDelete) parseRecordResults(ifc command, receiveSize int) 
 		}
 
 		// Aggregate metrics
-		metricsEnabled := cmd.node.cluster.metricsEnabled.Load()
+		metricsEnabled := cmd.node.cluster.metricsEnabled
 		if metricsEnabled {
 			cmd.node.stats.updateOrInsert(cmd.getNamespace(), cmd.getNamespaces(), cmd.commandType(), resultCode)
 		}
--- a/batch_command_exists.go
+++ b/batch_command_exists.go
@@ -111,7 +111,7 @@ func (cmd *batchCommandExists) parseRecordResults(ifc command, receiveSize int) 
 		}
 
 		// Aggregate metrics
-		metricsEnabled := cmd.node.cluster.metricsEnabled.Load()
+		metricsEnabled := cmd.node.cluster.metricsEnabled
 		if metricsEnabled {
 			cmd.node.stats.updateOrInsert(cmd.getNamespace(), cmd.getNamespaces(), cmd.commandType(), resultCode)
 		}
--- a/batch_command_get.go
+++ b/batch_command_get.go
@@ -147,7 +147,7 @@ func (cmd *batchCommandGet) parseRecordResults(ifc command, receiveSize int) (bo
 		}
 
 		// Aggregate metrics
-		metricsEnabled := cmd.node.cluster.metricsEnabled.Load()
+		metricsEnabled := cmd.node.cluster.metricsEnabled
 		if metricsEnabled {
 			cmd.node.stats.updateOrInsert(cmd.getNamespace(), cmd.getNamespaces(), cmd.commandType(), resultCode)
 		}
--- a/batch_command_operate.go
+++ b/batch_command_operate.go
@@ -164,7 +164,7 @@ func (cmd *batchCommandOperate) parseRecordResults(ifc command, receiveSize int)
 		}
 
 		// Aggregate metrics
-		metricsEnabled := cmd.node.cluster.metricsEnabled.Load()
+		metricsEnabled := cmd.node.cluster.metricsEnabled
 		if metricsEnabled {
 			cmd.node.stats.updateOrInsert(cmd.getNamespace(), cmd.getNamespaces(), cmd.commandType(), resultCode)
 		}
--- a/batch_command_udf.go
+++ b/batch_command_udf.go
@@ -105,7 +105,7 @@ func (cmd *batchCommandUDF) parseRecordResults(ifc command, receiveSize int) (bo
 		}
 
 		// Aggregate metrics
-		metricsEnabled := cmd.node.cluster.metricsEnabled.Load()
+		metricsEnabled := cmd.node.cluster.metricsEnabled
 		if metricsEnabled {
 			cmd.node.stats.updateOrInsert(cmd.getNamespace(), cmd.getNamespaces(), cmd.commandType(), resultCode)
 		}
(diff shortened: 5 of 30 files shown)


mirzakaracic:
@icellan thank you for the detailed write-up and for taking the time to investigate the regression

After reviewing the PR, we're unable to merge it as-is. We did find issues with some of the proposed changes — in particular, the concurrency-contract change around moving Cluster.metricsEnabled / metricsPolicy to plain (non-atomic) fields and the switch in internal/atomic.Int to a lock-free implementation have implications we're not comfortable accepting in this form, and the new ClientPolicy.MetricsPolicy field overlaps with mechanisms we want to keep consolidated.

That said, your central observation is valid and valuable: the hot-path helpers in command.go (applyTransactionErrorMetrics, applyTransactionRetryMetrics, applyConnectionRecoveredMetrics) should short-circuit when metrics are disabled, and the metrics-OFF path shouldn't be paying for work that ought to be zero-cost. That's a real regression and we want to fix it.

Our plan:
  - We'll close this PR.
  - We'll prepare our own fix targeting the hot-path gating in command.go, which will ship in the next Go client release.

We'll credit your investigation in the release notes. Thanks again for surfacing this.

PR #689, opened 2026-05-18
CLEINT-4753 CTX.mapKeysIn typed overloads diverge from CDT map-key spec

mirzakaracic:


cdt_context.go
```diff
@@ -271,6 +301,10 @@ func CtxMapIntKeysIn(keys ...int) *CDTContext {
 
 // CtxMapFloatKeysIn selects map entries whose keys are contained in the provided float keys.
 // Can be combined with CtxAndFilter to apply additional filtering on the selected entries.
+//
+// Deprecated: use CtxMapKeysIn with NewFloatValue per element, e.g.
+//
+//	CtxMapKeysIn(NewFloatValue(1.5), NewFloatValue(2.5))
```
ajkimdev: The CDT float map key type isn't supported by the Aerospike, so I'd recommend removing it entirely.  if removing it isn't an option, then at least make sure there are no examples that make it look like NewFloatValue() is usable.

khaf (approved):
Looks good.

PR #705, opened 2026-06-16
CLIENT-4975 Extended Error Handling changes

mirzakaracic:


evidence.yaml
```diff
@@ -0,0 +1,156 @@
+task_id: CLIENT-4975
```
khaf: Is this file necessary to be in the repo?

sub_code.go
```diff
@@ -0,0 +1,197 @@
+// Copyright 2014-2026 Aerospike, Inc.
```
khaf: This file should be in `types/result_code.go`.

sub_code.go
```diff
@@ -0,0 +1,197 @@
...
+// SubCodeNone (0) means "no subcode" - it is reserved universally and is the value
+// returned when the server did not send a subcode (verbosity disabled, or the
+// failing branch had no dispatchable subcode).
+//
+// This catalogue mirrors the server's per-status enums in as/include/base/proto.h
+// and is server-version-specific. It is append-only: published values are immutable
+// and are never renumbered or reused. New failure modes get new values appended to
+// their group. Treat any subcode value not declared here as an opaque integer
+// rather than assuming it is absent.
+const (
```
khaf: A `Stringer` interface implementation is also missing for this type.

sub_code.go
```diff
@@ -0,0 +1,197 @@
...
+	// SubCodeNone is returned when the server did not supply a subcode.
+	SubCodeNone = 0
```
khaf: `SubCode` should be a type so we can implement methods on it, including the `Stringer` interface. Refer to `ResultCode`.

khaf (changes requested):
There are minor things that require addressing.

khaf (dismissed):
Looks good.

PR #711, opened 2026-06-25
CLIENT-4940: change sendkey policy resolution to union of all send-key settings

dmishra-aero:
In a `batch`, consecutive same-namespace/set records can use a 1-byte "`repeat`" marker instead of a full header — but a record carrying its own user key can't repeat. The old code gated this on the parent `BatchPolicy.SendKey`, so enabling sendKey for writes in a mixed batch wrongly stripped the repeat flag from the reads too (reads never send a key). The fix gates on each record's **resolved** sendKey (reads = always false), so reads keep the repeat optimization in mixed write batches — smaller payloads, no behavior change.

### DIFF
--- a/batch_command_delete.go
+++ b/batch_command_delete.go
@@ -186,6 +186,10 @@ func (cmd *batchCommandDelete) commandType() commandType {
 
 func (cmd *batchCommandDelete) executeSingle(client *Client) Error {
 	policy := cmd.batchDeletePolicy.toWritePolicy(cmd.policy, client.dynConfig)
+	// Honor sendKey from BOTH the parent BatchPolicy and the per-record policy.
+	// toWritePolicy carries only the per-record value, so OR in the parent's sendKey here.
+	policy.SendKey = cmd.policy.SendKey || (cmd.batchDeletePolicy != nil && cmd.batchDeletePolicy.SendKey)
+
 	for _, offset := range cmd.batch.offsets {
 		key := cmd.keys[offset]
 		res, err := client.Operate(policy, key, DeleteOp())
--- a/batch_command_operate.go
+++ b/batch_command_operate.go
@@ -257,13 +257,18 @@ func (cmd *batchCommandOperate) executeSingle(client *Client) Error {
 			res, err = client.Operate(cmd.client.getUsableBatchReadPolicy(br.Policy).toWritePolicy(cmd.policy, client.dynConfig), br.Key, ops...)
 		case *BatchWrite:
 			policy := cmd.client.getUsableBatchWritePolicy(br.Policy).toWritePolicy(cmd.policy, client.dynConfig)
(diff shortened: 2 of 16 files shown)


BrianNichols (dismissed):

mirzakaracic (changes requested):
The base branch should be stage not v8.

mirzakaracic:
Few other observations:

1. Cluster default policy not unioned when a per-record policy is present

Java always ORs the cluster default into the resolved sendKey:
```
// BatchWrite.getSendKey
return writePolicyDefault.sendKey || (policy != null && policy.sendKey);
````

Go only consults the cluster default when the per-record policy is nil:
```
// resolveSendKey
wp := client.getUsableBatchWritePolicy(bw.Policy)
return parentPolicy.SendKey || (wp != nil && wp.SendKey)
getUsableBatchWritePolicy (client.go:2402) returns bw.Policy.patchDynamic(...) when bw.Policy != nil — it never folds in clnt.DefaultBatchWritePolicy.
```

Gap: DefaultBatchWritePolicy.SendKey=true, parent BatchPolicy.SendKey=false, and a record with its own BatchWritePolicy{SendKey:false} (set for an unrelated field like Expiration) → Java sends the key, Go drops it. Same for delete/UDF. This is the one behavioral divergence that could become a real bug report, and it isn't covered by the current tests. Suggest OR-ing clnt.DefaultBatch{Write,Delete,UDF}Policy.SendKey into resolveSendKey and adding a test for the non-nil-per-record + cluster-default-true case.

2. Dynamic-config sendKey: batch_write umbrella, Java, vs per-type keys, Go
I would confirmm what the intent is here. @BrianNichols thoughts?

Java's Command.resolveSendKey checks only dynamicConfiguration.dynamicBatchWriteConfig.sendKey for all record types — one dynamic key governs write, UDF, and delete. Go folds dynamic sendKey in per-type via patchDynamic, so batch_write, batch_delete, and batch_udf each gate their own type independently.

Result: dynamic batch_delete.send_key=true enables it on batch deletes in Go but is a no-op in Java.

dmishra-aero:
> For point 2: Java's Command.resolveSendKey checks only dynamicConfiguration.dynamicBatchWriteConfig.sendKey for all record types — one dynamic key governs write, UDF, and delete. Go folds dynamic sendKey in per-type via patchDynamic, so batch_write, batch_delete, and batch_udf each gate their own type independently.

This was the changes made : https://github.com/aerospike/aerospike-client-java/commit/9d73864ec8236d46f93348d079850ec33821225f

BrianNichols:
The initial java client commit only checked dynamic config "batch_write.send_key". The later java client commit restored the checks for "batch_delete.send_key" (for batch delete) and "batch_udf.send_key" (for batch UDF).

BrianNichols (approved):

mirzakaracic (approved):

PR #715, opened 2026-06-30
CLIENT-5000 fix issue where batch writes did not set per-record inDoubt on timeout (NO_RESPONSE)

mirzakaracic:


khaf (approved):
Looks good.

PR #717, opened 2026-07-07
CLIENT-4390 Create index to use “integer” instead of “numeric” starting with Aerospike server >= 8.1.3

mirzakaracic:


khaf (approved):
Looks good.

PR #718, opened 2026-07-07
CLIENT-5040 Fix ExpListAppendItems with a Map element throws PARAMETER_ERROR

mirzakaracic:


packer.go
```diff
@@ -128,6 +151,175 @@ func packIfcMap(cmd BufferEx, theMap map[any]any) (int, Error) {
...
+	canonicalRankList
+	canonicalRankMap
+	canonicalRankBytes
+	canonicalRankDouble
+	canonicalRankGeoJSON
+	canonicalRankOther
+)
+
+// canonicalKey is the normalized form of a map key used for canonical ordering.
+type canonicalKey struct {
```
khaf: This solution seems to be way too complex and can be done simpler with far less code. Let's have a meeting about it.

khaf (changes requested):

packer.go
khaf: Talked to @mirzakaracic , the algorithm currently allocates aggressively for sorting infrastructure, which seems to be unnecessary. We made the decision to simplify the code to reduce the allocations to only the key entries slice and nothing else.

packer.go
```diff
@@ -128,6 +152,214 @@ func packIfcMap(cmd BufferEx, theMap map[any]any) (int, Error) {
...
+		switch {
+		case aIsU && bIsU:
+			return cmp.Compare(au, bu)
+		case aIsU:
+			return 1
+		case bIsU:
+			return -1
+		}
+		return cmp.Compare(ai, bi)
+	case canonicalRankString, canonicalRankGeoJSON:
```
khaf, line 212: This line allocates. No need for `canonicalKeyString`. If the `case` is broken per type, the underlying values can be compared directly without the conversion/allocations.

packer.go
```diff
@@ -128,6 +152,214 @@ func packIfcMap(cmd BufferEx, theMap map[any]any) (int, Error) {
...
+		return cmp.Compare(canonicalKeyString(a), canonicalKeyString(b))
+	case canonicalRankBytes:
+		return bytes.Compare(canonicalKeyBytes(a), canonicalKeyBytes(b))
```
khaf: Same here.

packer.go
```diff
@@ -128,6 +152,214 @@ func packIfcMap(cmd BufferEx, theMap map[any]any) (int, Error) {
...
+		return int64(v), 0, false
+	case LongValue:
+		return int64(v), 0, false
+	case Value:
+		return canonicalKeyInt(v.GetObject())
+	}
+	return 0, 0, false
+}
+
+func canonicalKeyString(k any) string {
```
khaf: Should be removed.

packer.go
```diff
@@ -128,6 +152,214 @@ func packIfcMap(cmd BufferEx, theMap map[any]any) (int, Error) {
...
+		return string(v)
+	case Value:
+		if s, ok := v.GetObject().(string); ok {
+			return s
+		}
+	}
+	return ""
+}
+
+func canonicalKeyBytes(k any) []byte {
```
khaf: Also should be removed.

packer.go
```diff
@@ -128,6 +152,214 @@ func packIfcMap(cmd BufferEx, theMap map[any]any) (int, Error) {
...
+
+func canonicalKeyFloat(k any) float64 {
+	switch v := k.(type) {
+	case float32:
+		return float64(v)
+	case float64:
+		return v
+	case FloatValue:
+		return float64(v)
+	case Value:
```
khaf: Not sure this arm is needed. No other value object that returns float types.

packer.go
```diff
@@ -128,6 +152,214 @@ func packIfcMap(cmd BufferEx, theMap map[any]any) (int, Error) {
...
+	return ""
+}
+
+func canonicalKeyBytes(k any) []byte {
+	switch v := k.(type) {
+	case []byte:
+		return v
+	case BytesValue:
+		return v
+	case Value:
```
khaf: Not sure this arm is needed. No other value object that returns `[]byte`.

packer.go
```diff
@@ -128,6 +152,214 @@ func packIfcMap(cmd BufferEx, theMap map[any]any) (int, Error) {
...
+
+func canonicalKeyString(k any) string {
+	switch v := k.(type) {
+	case string:
+		return v
+	case StringValue:
+		return string(v)
+	case GeoJSONValue:
+		return string(v)
+	case Value:
```
khaf: Not sure this arm is needed. No other value object that returns `string` types.

khaf (changes requested):
Still a few little optimizations left on the table.

packer.go
```diff
@@ -128,6 +152,214 @@ func packIfcMap(cmd BufferEx, theMap map[any]any) (int, Error) {
 	return size, err
 }
 
+// canonicalPackBuffer marks a BufferEx as packing a filter-expression value
+// literal, which servers with AER-6930 require in canonical (key-ordered)
+// msgpack form. The marker travels down the recursive pack calls, so maps
+// nested anywhere inside the literal are packed in canonical order.
+type canonicalPackBuffer struct {
```
khaf: This cab done via:
```go
type canonicalPackBuffer BufferEx
```
To avoid allocations.

khaf (approved):
Looks good.

PR #719, opened 2026-07-08
CLIENT-4959 prevent txn abort

dmishra-aero:


mirzakaracic (changes requested):
Before this gets merged we need to double check with product how we want to proceed. But I have been looking at merged PR for [C#](https://github.com/aerospike/aerospike-client-csharp/pull/283). C# and Go have fundamental differences how they go about error handling however where C# and Go diverge now and where they have aligned before is state was set to committed immedietly after mark_roll_forward and before roll forward step. So after the PR the same failure mark succeeds, roll forrward fails would leave the transaction in different state for depending on what client you are looking at. In go the state would be committed and  in C# it is verified. Down stream this would drive different behavior. For instance a retry commit on verfied state would re-run the mark. In committed it would short circuit to already_committed.


Since @BrianNichols has alrready approved the PR for C# I would think C# has the correct behavior implemented. I would double check with product @tim-aero.

commit_status.go
```diff
@@ -21,7 +21,8 @@ const (
 	CommitStatusOK                   CommitStatus = "Commit succeeded"
 	CommitStatusUnverified           CommitStatus = "Commit process was disrupted on client side and unverified"
 	CommitStatusAlreadyCommitted     CommitStatus = "Already committed"
-	CommitStatusAlreadyAborted       CommitStatus = "Already aborted"
-	CommitStatusRollForwardAbandoned CommitStatus = "Transaction client roll forward abandoned. Server will eventually commit the Transaction."
+	CommitStatusAlreadyAborted           CommitStatus = "Already aborted"
```
mirzakaracic, line 24: nit: Fix formatting.

BrianNichols (commented):
The solution needs to be consistent across clients. Two tasks need to happen:

- @tim-aero Product should determine how the solution for the txn abort issue should work.
- Re-review each client to make sure it's solution is aligned with Product's solution.

mirzakaracic (approved):
Looks good.

abort_status.go
```diff
@@ -21,6 +21,7 @@ const (
 	AbortStatusOK                AbortStatus = "Abort succeeded"
 	AbortStatusAlreadyCommitted  AbortStatus = "Already committed"
 	AbortStatusAlreadyAborted    AbortStatus = "Already aborted"
+	AbortStatusCommitFailed      AbortStatus = "Abort not allowed because a commit already failed on this transaction with an in-doubt outcome"
```
gagan405, line 24: [nit] is  it customer visible message? There is a tiny mismatch with other messages - the other longer messages end with a `period` and this doesn't.

Ignore if thats not applicable here.

gagan405 (approved):
Approved with a minor cosmetic comment.

PR #739, opened 2026-07-31
CLIENT-5183 Reject negative selectByPath / modifyByPath flags instead of masking them into a valid operation

mirzakaracic:


khaf (approved):
Looks good.

PR #741, opened 2026-08-05
CLIENT-5198: parsing error fix for large compressed record for single read ops

dmishra-aero:


mirzakaracic (changes requested):
I would add an integration test here. Not sure how feasable that is but the reason we did not catch this to begin with is because we don't have code that exercises this code. The unit test case is good but will not catch issue in the future since you have fixed fixtures.

dmishra-aero:
changes addressed..

mirzakaracic (approved):

PR #743, opened 2026-08-12
Minor fixes

khaf:
WIP, don't merge please


marshal_test.go
```diff
@@ -0,0 +1,72 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package aerospike
+
+import (
+	"reflect"
```
mirzakaracic, line 18: issue: getting a failure when we on build now when we try to run go vet for getMapping(....)

command.go
```diff
@@ -3934,16 +3763,33 @@ func (cmd *baseCommand) executeAt(ifc command, policy *BasePolicy, deadline time
...
+			// breaker, matching the Java client. The response was fully
+			// parsed, so the connection is healthy and goes back into the
+			// pool rather than being closed. (Multi commands never pool on
+			// error -- canPutConnBack -- since their stream may be undrained;
+			// they close and retry through the batch machinery instead.)
+			if (overloadedServerError(err) || parsedServerTimeout(err)) && !cmd.oneShot {
+				isClientTimeout = false
+				cmd.node.incrErrorCount()
+
+				if ifc.canPutConnBack() && cmd.conn.IsConnected() {
```
mirzakaracic, line 3779: issue: not checking type of error we could potentially try to put connection back that should not be put back. 

We have KeepConnection functions we should use

command.go
```diff
@@ -4038,17 +3864,49 @@ func (cmd *baseCommand) parseRecordResults(ifc command, receiveSize int) (bool,
...
+	return !errors.As(err, &netErr)
+}
+
+// overloadedServerError reports the transient server responses that are
+// retried and feed the circuit breaker, matching the Java client: the node is
+// momentarily overloaded (DEVICE_OVERLOAD) or the record is locked by another
+// transaction (KEY_BUSY). Both arrive as fully parsed responses on a healthy
+// connection.
+func overloadedServerError(err Error) bool {
+	return err.Matches(types.DEVICE_OVERLOAD, types.KEY_BUSY)
```
mirzakaracic, line 3904: question: we seem to be treating DEVICE_OVERLOADED and KEY_ERROR the same in this case. Is this intentional?

command.go
```diff
@@ -3918,6 +3743,10 @@ func (cmd *baseCommand) executeAt(ifc command, policy *BasePolicy, deadline time
 			continue
 		}
 
+		// The command reached the wire: from here on a failed write may have
+		// been applied by the server, which is what inDoubt reports.
+		cmd.commandSentCounter++
```
mirzakaracic, line 3748: question: I can see the comment. I guess is to follow in suit with other clients but is this inentional? The reason I am asking is because this does affect MaxRetries.

command.go
```diff
@@ -3563,15 +3381,13 @@ func (cmd *baseCommand) sizeBufferSz(size int, willCompress bool) Error {
...
-	if size <= len(cmd.dataBuffer) {
-		// don't touch the buffer
-		// this is a noop, here to silence the linters
-		cmd.dataBuffer = cmd.dataBuffer
-	} else if size <= cap(cmd.dataBuffer) {
-		cmd.dataBuffer = cmd.dataBuffer[:size]
-	} else {
+	// Grow the buffer only when needed; a buffer at least `size` long is left
+	// untouched.
+	if size > cap(cmd.dataBuffer) {
```
mirzakaracic, line 3386: issue: The logic here not same. I would not drink AI's coolaid on this one.

mirzakaracic (changes requested):
Most of the concerns I have are in the command.go there. I think the metrics changes are good but I would recommend running bench_apply_transaction_metrics.go before and after the changes just to make we have not digressed somehow.

PR #746, opened 2026-08-18
CLIENT-5260 Align TxnState/COMMIT_FAILED enum ordering with C client

dmishra-aero:


dmishra-aero:
Git Action are failing here and being addressed here in this [PR](https://github.com/aerospike/aerospike-client-go/pull/747) so we merge that before this one goes

mirzakaracic (approved):

khaf:
@mirzakaracic This is a major breaking change. Should be kept until a major release version.

ps. Talked to @mirzakaracic and turns out I was wrong. The change was introduced recently and itself was a breaking change.

PR #762, opened 2026-08-27
CLIENT-5361 Expose CREATE_ONLY and UPDATE_ONLY string write flags

mirzakaracic:
If testing locally for all tests to pass locally you will have to have changes in CLIENT-5308


khaf (approved):
Looks good.

PR #764, opened 2026-08-28
CLIENT-5116 Add ExpFromAEL(text) — name the AEL-text wire primitive the Go client already carries

mirzakaracic:


khaf (approved):
Looks good.

PR #767, opened 2026-09-01
CLIENT-5267: Harden dynconfig sendKey test for multi-node clusters

dmishra-aero:


batch_test.go
```diff
@@ -874,6 +880,19 @@ var _ = gg.Describe("Aerospike", func() {
 				}
 
 				if nsInfo(ns, "storage-engine") == "device" {
+					// Multi-node skip (temporary).
+					//
+					// Expected: per-key errors (RECORD_TOO_BIG, INVALID_NAMESPACE) are reported on
+					// each BatchRec; BatchOperate returns err == nil (same idea as Java operate():
+					// false status, inspect resultCode — no exception).
+					//
+					// Today on multi-node, executeSingle and some batch-wire responses can also
```
dmishra-aero: @mirzakaracic This test we need to review or fix separately when run in multi node set up, i guess this test is flaky but i feel it might need client adjustment to meet up with test expectation

.github/workflows/build-multi-node.yml
```diff
@@ -27,10 +29,10 @@ permissions:
 
 env:
   # defaults to use against workflow triggered through cron
-  AEROSPIKE_VERSION: 8.1.1
+  AEROSPIKE_VERSION: 8.1.1.0
   NODES: 3
   RUNNER_VERSION: ubuntu-24.04
-  MODES: AP,SC
```
mirzakaracic, line 33: question: why remove SC mode testing?

.github/workflows/build-multi-node.yml
```diff
@@ -78,6 +80,7 @@ jobs:
       id-token: write
       actions: write 
     strategy:
+      fail-fast: false
```
mirzakaracic: issue: there is no reason to have other forks continue if one fails. Either all are good or none.

dmishra-aero: This i will revert..anyway.. I added initially testing sake to at-least check what test's are failing on both branches.

.github/workflows/build-multi-node.yml
dmishra-aero: I guess this workflow wasn’t running automatically on PRs earlier and could only be triggered manually. I added this so that we have at least one leg running for the AP namespaced multi-node cluster.

Anyway we could run for SC also or during PR
Also one observation lot of test are failing in SC for changes like CDT, String ops in SC mode when i ran for multi node cluster ..that we need to separately take a look

batch_test.go
```diff
@@ -874,6 +880,19 @@ var _ = gg.Describe("Aerospike", func() {
 				}
 
 				if nsInfo(ns, "storage-engine") == "device" {
+					// Multi-node skip (temporary).
+					//
+					// Expected: per-key errors (RECORD_TOO_BIG, INVALID_NAMESPACE) are reported on
```
mirzakaracic: issue: so this is a smell to me because this is telling me we have divergent behavior and this is a stop gap measure to put on band aid on a hole. 

We should fix the issue and not relax contraints. Remove this.

dmishra-aero: yeah this is what i reffered [here](https://github.com/aerospike/aerospike-client-go/pull/767#discussion_r3906165513) ?
I am not sure what is ideal expectation here .

mirzakaracic:
Reviewed the CI changes here in detail. One blocking defect that's unrelated to the stated purpose of the PR, plus two coverage concerns. The batch skip itself is now tracked as **CLIENT-5411**.

---

### 1. Blocking: `aerospike-version` is not a valid input in the new action version — CI is silently running `:latest`

This PR bumps all four workflows to `setup-aerospike-server@0d843cff` (v4.1.0), which renamed `aerospike-version` to `server-tag`. The rename was applied in only one of them:

| file | input | status |
|---|---|---|
| `build-multi-node.yml:97` | `server-tag:` | correct |
| `build.yml:95` | `aerospike-version:` | **ignored** |
| `nightly.yml:103` | `aerospike-version:` | **ignored** |
| `build-nightly.yml:101` | `aerospike-version:` | **ignored** |

From this PR's own single-node job log (`build (8.1.2.0, 8.1.2.0, 1.23.0, AP)`):

```
##[warning]Unexpected input(s) 'aerospike-version', valid inputs are ['server-tag', 'num-nodes', ...]
  aerospike-version: 8.1.2.0
  server-tag: latest
...
Aerospike server image: aerospike.jfrog.io/database-docker-virtual/aerospike-server-enterprise:latest
Aerospike server version: Aerospike Enterprise Edition build 8.1.2.4-4
```

The job advertising `8.1.2.0` actually tested **8.1.2.4-4** off the floating `:latest` tag. Actions warns but does not fail on an unexpected composite-action input, so this passes green. Net effect: the server-version matrix is non-functional for all single-node and nightly CI, and the check names, coverage artifacts and PR statuses report a version that was never exercised. It will keep drifting as `latest` moves.

Please change these three to `server-tag:`.

### 2. `MODES: AP,SC` -> `MODES: AP` removes all multi-node SC coverage

`env.MODES` governs the `pull_request` and cron triggers (the `workflow_dispatch` default stays `AP,SC`), so as shipped, automated multi-node runs are AP-only. Confirmed by this PR's checks — multi-node produced a single job, `build (stable, 8.1.1.0, 1.23.0, AP)`.

This interacts with the skip in `batch_test.go` in a way that's worth being explicit about. In the pinned action, device storage exists **only** in SC mode (`action.yaml:528-534`: SC -> `namespace-sc.conf` with `storage-engine device`; AP -> `namespace-memory.conf` with `storage-engine memory`). So:

- PR / cron multi-node (AP) -> memory namespace -> the `storage-engine == device` block never runs at all
- Manual dispatch (AP,SC) -> device namespace -> block runs -> the new node-count skip fires
- `build.yml` single-node SC -> device -> block runs on 1 node -> passes

The two changes therefore overlap to close every path: dropping SC removes the only automated trigger that could reach the failing assertions, and the skip covers the manual trigger that remains. Cron multi-node SC is presumably what surfaced the failure in the first place. Could you either restore `AP,SC` or state in the PR description why multi-node SC is being dropped? Right now the description is empty and this reads as an unintended side effect of the version-format fix.

### 3. Skip scoping in `batch_test.go`

The underlying issue is real and confirmed (CLIENT-5411): on a batch node holding a single key, `executeSingle` returns server row result codes as a top-level error, while the multi-key wire path reports them per-record. Reproduces deterministically on a **single** node with a 1-record `BatchOperate`, and Java takes the same `offsetsSize == 1` fast path without throwing. So skipping is defensible as a temporary unblock, with two asks:

- Reference CLIENT-5411 in the skip message. "until BatchOperate row-error propagation is finalized" names no ticket, so nothing tracks it.
- Gate on the ticket, not on `len(client.GetNodes()) > 1`. Node count is a symptom; the trigger is offsets-per-node. A future one-record batch test would hit this on single-node CI with no guard. Splitting the `RECORD_TOO_BIG` assertions into their own `It` would also keep the `INVALID_NAMESPACE` row assertion running everywhere — that one passes on multi-node today and is currently lost as collateral.

### Looks good

- The `nsupPeriod(ns) == 0` skip on `BatchRead TTL` is correct. Neither namespace template in the action sets `nsup-period`, so it defaults to 0 and `allow-ttl-without-nsup` is false — the skip will legitimately fire in CI.
- `dynconfig_serialze_test.go` readability pass and the `NewClientWithPolicyAndHost` fix in `utf8_validate_test.go` are clear improvements.

### Minor

`aerospike_suite_test.go` still has two redundant `if isSCMode { defaultBatchPolicy.ReadModeSC = ... }` blocks (lines ~137 and ~143) and two duplicated `SetDefaultBatchPolicy` / `SetDefaultWritePolicy` calls. Pre-existing, and the copy-paste target was probably meant to be `defaultScanPolicy` / `defaultQueryPolicy`, which still never get `ReadModeSC` in SC mode. Since you're already in this function, might be worth cleaning up — though it's moot while `MODES: AP`, which is itself part of why it stayed invisible.

khaf (approved):
Looks good.

khaf:
@mirzakaracic I'll let you merge this one when ready.

PR #772, opened 2026-09-05
CLIENT-5411 Keep server row errors on BatchRecord in executeSingle path

dmishra-aero:
The change is focused on **`BatchRecord` APIs** (`BatchOperate`, etc.), where server-side row errors should remain on `BatchRecord.Err` rather than becoming the top-level `err` from the 1-key `executeSingle` path.

**Note:** 
`BatchGet` and `BatchExists` are different as they return `[]*Record` and `[]bool` without a per-key error field. Changing their behavior would turn a real row failure into `err == nil` with `nil`/`false`, making it indistinguishable from a missing key.

For callers that need per-key error details, **`BatchGetComplex`** provides `BatchRecord.Err` and remains covered by `batch_index_command_get.go`.

So this PR is intentionally scoped to the **`BatchRecord`-based APIs**, avoiding an untested behavior change to the simpler read APIs that have no way to represent row-level errors.

### DIFF
--- a/error.go
+++ b/error.go
@@ -17,6 +17,7 @@ package aerospike
 import (
 	"errors"
 	"fmt"
+	"net"
 	"runtime"
 	"strconv"
 	"strings"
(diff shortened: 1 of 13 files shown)


error.go
```diff
@@ -186,6 +187,57 @@ func newServerError(code types.ResultCode, serverMessage string, subcode types.S
...
 }
 
+// isClientTimeout reports whether a TIMEOUT error originated on the client side
+// (policy deadline, socket deadline, or connection deadline), as opposed to a
+// server protocol result code.
+func isClientTimeout(err Error) bool {
+	errTimeoutMsg := ErrTimeout.msg
+	errNetTimeoutMsg := ErrNetTimeout.msg
+
+	for e := error(err); e != nil; {
```
mirzakaracic: Issue: you don't have to enumerate all the errors that are there. Should be able to check the last appended error and be done with it.

error.go
```diff
@@ -186,6 +187,57 @@ func newServerError(code types.ResultCode, serverMessage string, subcode types.S
...
+}
+
+// shouldAbortBatchCommand reports whether err from executeSingle should fail
+// the whole batch subcommand instead of remaining a per-record row error.
+func shouldAbortBatchCommand(err Error) bool {
+	if err == nil {
+		return false
+	}
+	code := err.resultCode()
+	if code < 0 {
```
mirzakaracic: issue: Not all postitive error cods are comming from the se server

invalid namespace or code 20 or parameter_error or code 4 could be coming from the client.

error.go
```diff
@@ -186,6 +187,57 @@ func newServerError(code types.ResultCode, serverMessage string, subcode types.S
 	return ne
 }
 
+// isClientTimeout reports whether a TIMEOUT error originated on the client side
+// (policy deadline, socket deadline, or connection deadline), as opposed to a
+// server protocol result code.
+func isClientTimeout(err Error) bool {
+	errTimeoutMsg := ErrTimeout.msg
+	errNetTimeoutMsg := ErrNetTimeout.msg
```
mirzakaracic: issue: using message to check for equlity looks like something that could shoot us in the foot in the future.

mirzakaracic (changes requested):
I would move create a single flag in the AeroSprikeError that would allow us to check on error creation if response/error is coming from the server. You could then do something like 

```
  func isServerError(err Error) bool {
        switch e := err.(type) {
        case *AerospikeError:
                return e.fromServer
        case *constAerospikeError:
                return e.fromServer
        }
        return false
  }

  func shouldAbortBatchCommand(err Error) bool {
        return err != nil && !isServerError(err)
  }
```

error.go
dmishra-aero: `message compare` was a workaround because `errors.Is(err, ErrTimeout)` matches plain server TIMEOUT. will remove once errors source is embedded into `Aerospike Error struct`

error.go
dmishra-aero: `client timeouts` usually sit on the outer link via `chainErrors`, but some paths (e.g. wrapped os.ErrDeadlineExceeded) still need chain inspection unless we tag at creation.

dmishra-aero:
> I would move create a single flag in the AeroSprikeError that would allow us to check on error creation if response/error is coming from the server. You could then do something like

`fromServer` is intentionally not copied in `As()`: `isServerError` uses a type switch rather than `errors.As`, and `chainErrors` already copies the entire struct by `value`, including `fromServer`.

mirzakaracic (approved):

PR #774, opened 2026-09-10
CLIENT-5442  regex_replace drops the StringPolicy write flags on both the operate and expression paths

mirzakaracic:


khaf (approved):
Looks good.

PR #790, opened 2026-09-21
CLIENT-5514 connection buffer is returned to the shared pool while still in use when UseCompression is enabled

mirzakaracic:


Voyagerroc-Lab (commented):
Reviewed the executeAt buffer lifetime fix and the new threshold-focused regression harness. Removing pool return at command completion matches the connection-owned buffer lifetime, while the test exercises both compressed/uncompressed paths and snapshots buffer aliases before the connection is returned. The multi-node and client test matrices are running successfully. I found no blocking issue.

